PR #58, opened 2024-07-22
Fix dependencies and override deprecated URL functions

edigerad:
Fix dependencies and override deprecated URL functions

*  Updated dependencies in requirements.txt: fixed versions for 'django' and added 'djangorestframework==3.15.2'.
*  Replaced deprecated 'url' functions with 'path' and 're_path' in tests/urls.py for compatibility with modern Django versions.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -50,4 +50,4 @@ CACHES = {
 #     'contenttypes': None,
 # }
 
-ALLOWED_HOSTS = ['localhost', '127.0.0.1', '*']
+ALLOWED_HOSTS = ['localhost', '127.0.0.1']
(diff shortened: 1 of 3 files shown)


tests/settings.py
```diff
@@ -46,4 +48,6 @@
 # MIGRATION_MODULES = {
 #     'auth': None,
 #     'contenttypes': None,
-# }
\ No newline at end of file
+# }
+
+ALLOWED_HOSTS = ['localhost', '127.0.0.1', '*']
```
Sultan91, line 53: Is `"*"` really required?

dbtiunov (approved):

PR #77, opened 2026-04-17
Release/0.3.0

emil-balashov:
> [!NOTE]
> **High Risk**
> Large breaking release touching core transition execution, new PostgreSQL-backed background durability, and required Celery/beat/broker configuration for production correctness.
> 
> **Overview**
> **Release 0.3.0** ships a first-class **`django_logic.background`** layer: `BackgroundTransition` / `BackgroundAction`, the durable `TransitionMessage` model (migrations, partial-unique concurrency, timing/timeouts), plus dispatch, phase-2 runner, Celery tasks, and four periodic safety-net jobs. Sync mode (`BACKGROUND_EXECUTION='sync'`, `sync_execution()`, `retry_pending()`) is added for tests without a broker.
> 
> The synchronous core is reworked around that engine: **failure side-effects**, structured **`django-logic` / `django-logic.transition`** logging (legacy `LogType` / custom logger removed), **`Process`** validation (unique `in_progress_state` and background `action_name`s), transition **context propagation** (`tr_id` / `root_id`), and **`State` / `RedisState`** locking and persistence fixes. **Celery, Redis, and DRF** move to optional extras; the in-tree **`demo/`** app and **Travis** are removed.
> 
> **DX & ops:** large **README** background/deployment section (beat, `task_reject_on_worker_lost`, pgbouncer), **GitHub Actions** CI (multi-Python) and **stability** workflow (Postgres + Redis), **Docker** / compose for stability tests, **`django_logic.conditions`** helpers, observability (**Celery `shadow`**, optional Sentry transaction naming), and AI-oriented **`.cursor` / `CLAUDE.md`** guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5559232c5f826957283ba220ef131d611d1c4f24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


emil-balashov:
## Follow-up: phase-2 timing + dead-code cleanup (04d885a)

Two themes in one push.

### 1. TransitionMessage timing fields

New on the model, migration `0002_transitionmessage_timing`:

- `started_at` — written at the top of every phase-2 attempt (overwritten on retry, so a watchdog sees the *current* attempt)
- `completed_at` — written once when the row is marked completed (success or terminal failure)
- `duration_ms` — computed on completion; clock-skew-clamped to ≥ 0
- Index on `(is_completed, started_at)` — what the future watchdog query needs

This unblocks the remaining S1-tier reliability work (per-transition timeout / watchdog) and the AI-readable test timeline output planned for stage 3.

### 2. Bug fix: unrestorable TMs were looping forever

The `_Restore_Failed` path logged *"Marking completed to stop retries"* but called `mark_as_completed()` inside the atomic block. The rollback silently undid it — the row stayed uncompleted, `errors_count` never incremented, and the periodic starter re-dispatched it on every tick.

Fix: `_RestoreFailed` now propagates out of `_run_atomic`; `_mark_unrestorable_completed(tm_id)` runs a single `UPDATE` outside the rolled-back transaction. The new invariant comment at `runner.py:124-129` codifies "mark_as_* inside the atomic; `_RestoreFailed` is the documented escape."

### 3. Dead-code cleanup (pre-1.0, no backward-compat constraint)

Deleted outright — zero production callers, only their own tests kept them alive:

| Removed | Replaced by |
|---|---|
| `django_logic/utils.py` (`restore_user_object`, `restore_action`, `get_process_instance`, `get_process_and_state`) | `background/runner.py::_restore` + `background/serializers.py::restore_user` |
| `Transition.get_task_kwargs` | `background/serializers.py::serialize_kwargs` |
| `ProcessManager.bind_state_fields` / `non_state_fields` / `save()` | `ProcessManager.bind_model_process` |
| `TransitionEventType.BACKGROUND_MODE` + `_BACKGROUND_MODE_KEYS` filter | — (pre-0.3.0 fire-and-forget remnant) |
| `State.get_log_data` | — (no callers) |
| `queryset_name` plumbing on `Process` and `State` | `model._default_manager` |
| `ignore_sources` parameter on `Process.get_available_transitions` / `get_transition_by_action_name` | `in_progress_state` is already appended to `sources` in `Transition.__init__` |

Behavior tweak worth reviewing: `State.set_state` now calls `refresh_from_db(fields=[self.field_name])` instead of a full refresh. Side-effect mutations to other attributes survive the state write — previously they were silently reverted in-memory.

### Tests

193 pass (from 207). 14 removed tests were testing deleted code; 4 new timing tests added in `tests/background/test_timing.py`.

### Net diff

`django_logic/` core: **882 → 748 LOC** (-15%). Total: **+384 / -483** across 19 files.

emil-balashov:
## Stable background execution — 5 reliability guarantees (586aad2)

Closes the S1 reliability gaps called out in the earlier plan review. Every in-flight `BackgroundTransition` row now has a guaranteed terminal exit path; no "stuck forever" state remains.

### 1. Phase-2 restore is unambiguous
Previously `_find_transition` matched on `action_name` and took the first hit — the docstring hand-waved the ambiguity as "caller's bug." Now rejected at class-creation time:

- New `is_background` marker on `BaseTransition` (overridden in `BackgroundTransition`) — avoids a circular import at validation time.
- New `_validate_unique_background_action_names(process_cls)` in `Process.__init_subclass__`:
  - Two background transitions sharing an `action_name` → `ImproperlyConfigured`.
  - A synchronous `Transition` colliding with a background one → `ImproperlyConfigured`.
- Sync-only duplicate action_names are still allowed (the sync call path's `get_transition_by_action_name` disambiguates via conditions/permissions).

### 2. Celery-mode + SQLite rejected at `validate_on_ready`
`select_for_update(nowait=True)` is a no-op on SQLite and partial unique indexes aren't supported, so the phase-2 concurrency guard silently degrades to "serialize everything." Misconfig now fails loud at app-ready.

### 3. Swallowed `failure_side_effects` exceptions are visible
- New `failure_side_effect_error: TextField` on `TransitionMessage` (migration `0003`).
- `FailureSideEffects.execute` returns the swallowed exception (or `None`).
- `_handle_failure` records it via `tm.record_failure_side_effect_error`. Broken cleanup is no longer invisible.

### 4. `detect_stuck_transitions` terminates, not just logs
Rows stuck at `errors_count >= MAX_ERRORS` are forcibly finalized:
- Row-locked via `select_for_update(nowait=True)` — rows held by a live worker are deferred.
- `failed_state` written, `failure_side_effects` run, marked completed.
- Unrestorable rows (model/transition renamed) are still marked completed so the retry loop stops.

### 5. Per-attempt timeout watchdog
- `BackgroundTransition(timeout=N)` — positive int, seconds, opt-in (None = no watchdog).
- `timeout_seconds: PositiveIntegerField` on TM (migration `0004`), populated at phase 1.
- New `watchdog_stale_attempts` periodic task: scans uncompleted rows with `started_at + timeout_seconds < now`, records a synthetic `TimeoutError`, and at `MAX_ERRORS` finalizes in the same atomic block (same row lock, no recursion/deadlock).
- DB-side scan floor: `Min(timeout_seconds)` narrows the candidate set before the Python per-row comparison — bounded cost at scale.

### Shared terminal path
`_finalize_terminal_from_watchdog(tm, exception, source)` factors out `failed_state` + `failure_side_effects` + `mark_as_completed` so the watchdog and detect_stuck terminate rows the same way the in-phase-2 failure path does. Previously my first cut skipped `failure_side_effects` in those paths — that's now symmetric.

### Idempotency caveat, explicit
`abandon_timed_out_attempt` docstring spells out the "live-but-slow worker that dropped its row lock" edge case: the watchdog cannot distinguish an abandoned attempt from a slow-but-alive one that has released its `select_for_update`. Safe per the reliability contract (§2.7 — side-effects must be idempotent). The original worker's eventual `mark_as_completed` / `record_error` will fail harmlessly against a completed row.

### Test infrastructure
- `tests/settings.py` defaults to SQLite (zero setup for contributors), switches to Postgres when `POSTGRES_HOST` is set. Sync mode doesn't exercise real `select_for_update` concurrency, so SQLite is safe for unit tests.
- Stability runs opt into Postgres via env var.
- **217 tests pass on both** (SQLite default and `POSTGRES_HOST=localhost`). 9 skipped — all `@requires_real_redis`.

### New tests
- `tests/background/test_validation.py::UniqueBackgroundActionNameTests` — 4 tests
- `tests/background/test_failure_side_effects.py` — 2 tests
- `tests/background/test_watchdog.py` — 14 tests (kwarg validation, phase-1 persistence, watchdog skip paths, watchdog act paths)
- `tests/background/test_safety_net.py::DetectStuckTests` rewritten — 4 tests for the new terminate behavior
- `tests/background/test_sync_execution.py::CeleryModeWithoutCeleryTests` — 1 test for the SQLite guard

### Reliability posture now

Terminal exits for any phase-2 row:
1. Normal success → `target`, `mark_as_completed`.
2. In-phase-2 failure at MAX_ERRORS → `failed_state` + `failure_side_effects` + `mark_as_completed`.
3. `retry_stale_transitions` re-dispatch eventually hits MAX_ERRORS via (2).
4. `detect_stuck_transitions` force-finalizes MAX_ERRORS rows missed by (2) (e.g., worker killed mid-atomic).
5. `watchdog_stale_attempts` kills timed-out attempts by recording a TimeoutError and, at MAX_ERRORS, finalizing.
6. `_RestoreFailed` marks the row completed outside the rolled-back atomic (already in from the previous commit).

### Net diff

968 insertions, 46 deletions across 18 files. 4 new test files. 2 new migrations (`0003`, `0004`).

PR #99, opened 2026-06-15
feat(background): support condition-disambiguated nested background transitions (#98)

### LINKED ISSUE #98: Support condition-disambiguated background transitions that share an action_name across nested processes
## Summary

`BackgroundTransition` / `BackgroundAction` instances may not share an `action_name` across a `Process` and its `nested_processes` — `Process.__init_subclass__` → `_validate_unique_background_action_names` raises `ImproperlyConfigured`. This blocks the **condition-disambiguated nested-process pattern** (polymorphic routing by a condition on the instance) for background transitions, even though the **synchronous** path already supports it via `get_transition_by_action_name` (which disambiguates on conditions/permissions at runtime).

The restriction exists only because phase-2 restore (`background/runner.py::_find_transition`) looks the transition up by `TransitionMessage.transition_name` (= `action_name`) **alone** — first match while descending `nested_processes` — and deliberately bypasses the condition/source gating. The `TransitionMessage` carries no discriminator for *which* nested transition was enqueued.

## Affected code

- `django_logic/process.py` — `_validate_unique_background_action_names` (raises the error; explicitly documents that `action_name` is the only restore key).
- `django_logic/background/runner.py` — `_find_transition` (first-match-by-`action_name`, recursing into `nested_processes`; comment notes it bypasses the state/condition gating because phase 2 runs while the instance sits in `in_progress_state`, not in the transition's `sources`).
- `django_logic/background/models.py` — `TransitionMessage` stores `app_label`, `model_name`, `instance_id`, `process_name`, `field_name`, `transition_name`, `queue_name` — nothing that distinguishes sibling nested transitions sharing an `action_name`.

## Scenario

A `ConversationProcess` routes per messaging integration via two nested processes — `GmailConversationProcess` and `DummyConversationProcess` — each declaring a background transition `send_message_via_integration` (open→open) and `close` (open→closed), selected by a condition on the instance (`source_integration == gmail` / `dummy`). This is exactly the polymorphism the sync path supports (e.g. courier-style sub-processes). At class creation it fails:

```
ImproperlyConfigured: Process ConversationProcess (or its nested processes) has two
background transitions sharing action_name='send_message_via_integration' … background
action_names must be unique across a Process and its nested processes.
```

There is no way to keep the per-integration nested structure **and** make these transitions durable/background — you must either give each integration a distinct `action_name` (which breaks the generic callers that just call `conversation.process.send_message_via_integration(...)`) or drop durability.

## Suggested fix

Make phase-2 restore able to identify the exact background transition, then relax `_validate_unique_background_action_names` to allow duplicates that are condition-disambiguated. Two viable approaches:

1. **Record the resolved transition at phase 1 (preferred — deterministic).** Phase 1 already resolves the exact transition (it evaluates sources/conditions *before* moving the instance to `in_progress_state`). Persist a discriminator on `TransitionMessage` — e.g. the owning (nested) process class dotted path, or a stable transition key — alongside `transition_name`. `_find_transition` uses it to select the exact transition; rows without it fall back to the current first-match behaviour (backward compatible). The validator then only forbids genuinely ambiguous duplicates.

2. **Re-evaluate conditions at restore.** Have `_find_transition` collect all `action_name` matches across the tree and pick the one whose `conditions` pass for the instance (skipping the `sources` membership check, since the instance is mid-transition in `in_progress_state`). Simpler, but only correct when conditions alone disambiguate and don't depend on the pre-transition source state.

Approach 1 avoids relying on conditions being re-evaluable mid-transition and is robust to overlapping/empty conditions.

## Found in

django-logic `master` (the 0.4.0 line vendored downstream), while migrating a downstream `conversations_app` onto the engine. The conversation integration uses condition-disambiguated nested Gmail/Dummy processes with background `send_message_via_integration` / `close`.

emil-balashov:
Closes #98.

## Problem

`BackgroundTransition` / `BackgroundAction` could not share an `action_name` across a `Process` and its `nested_processes` — `_validate_unique_background_action_names` raised `ImproperlyConfigured`. This blocked the **condition-disambiguated nested-process pattern** (polymorphic routing by a condition on the instance) for background transitions, even though the **synchronous** path already supports it.

The restriction existed only because phase-2 restore (`runner._find_transition`) looked the transition up by `action_name` alone (first match descending `nested_processes`) and had no discriminator for *which* condition-disambiguated sibling phase 1 had enqueued.

## Solution (Approach 1 from the issue — deterministic)

Phase 1 already resolves the exact transition (conditions/permissions evaluated before the instance moves to `in_progress_state`). It now **records the owning (nested) process class** on the `TransitionMessage`, and phase 2 restores that exact transition from it — without re-evaluating the condition, so routing stays deterministic even if the instance changes mid-flight.

- **`process.py`** — `_get_transition_method` resolves `(transition, owning_process)` and records `owning_process_class` for background transitions (overwrite, never inherited via a chained `next_transition`). New `_iter_available_with_owner` / `_resolve_transition_with_owner`; `get_available_transitions` / `get_transition_by_action_name` delegate to them (public behaviour unchanged).
- **Validator relaxed** — `_validate_unique_background_action_names` now forbids only the genuinely ambiguous case (two background transitions sharing a name **within a single process class**, where `(owning class, action_name)` no longer identifies one transition). Duplicates across **distinct** nested classes are allowed. The background-vs-synchronous name-collision check is **unchanged**.
- **`runner._find_transition`** — selects via the recorded owner (`is_background`-filtered). Rows with a blank owner (created before this change, or whose transition lives on the bound process itself) fall back to the previous first-match-by-`action_name` behaviour; a recorded owner no longer in the tree (renamed/removed between deploys) logs a warning and degrades to first-match rather than stranding the instance.
- **`serializers.serialize_kwargs`** — strips `owning_process_class` from the persisted kwargs JSON so it stays column-only and never reaches side-effect/callback kwargs (engine bookkeeping, not caller data).
- **`models.py` + migration `0007`** — new `owning_process_class` column (blank default — backward compatible). `transitions._phase_one_atomic` writes it. `testing/snapshot.py` round-trips it.

## Backward compatibility

Fully backward compatible. Legacy `TransitionMessage` rows (blank `owning_process_class`) resolve via the unchanged first-match path; the old validator guaranteed that match was unique for them. Migration `0007` is an additive nullable/blank column — no backfill, in-flight rows survive (default `''`).

## Tests

- `tests/background/test_condition_disambiguated_nested.py` (new): end-to-end Gmail/Dummy routing; a clincher proving phase 2 does **not** re-evaluate the condition (Approach 1 vs Approach 2); blank-owner and unknown-owner first-match fallbacks; phase-1 ambiguity (overlapping conditions) raising cleanly **before** any `in_progress_state` write or `TransitionMessage` row.
- `Conversation` + `Gmail`/`Dummy` + overlapping-condition fixtures (no new schema beyond the `Conversation` model, migration `0004`).
- `test_kwargs_roundtrip.py`: pins that `owning_process_class` is absent from phase-2 side-effect kwargs (bound **and** nested owner).
- `test_validation.py`: updated for the relaxed rule (cross-class duplication allowed; within-class still rejected).

**347 tests pass** (+14); `makemigrations --check` clean; migration `0007` applies on a fresh DB.

## Review

Verified via a 4-dimension adversarial review (backward-compat/migration, owner-threading, durability/concurrency, validator/tests) with per-finding verification: **0 correctness defects**; the 2 confirmed findings were test-coverage gaps, both addressed in this PR.

## Note for reviewers

The validator keeps the **background-vs-synchronous** name-collision rejection (conservative, in scope). The fully-principled alternative would drop it too — phase 2 already filters to `is_background` — but that broadens behaviour beyond the issue; happy to follow up separately if preferred.

---

> [!NOTE]
> **Medium Risk**
> Changes phase-2 transition resolution and adds a migration on the hot `transitionmessage` table; legacy in-flight rows can finalize without running work if a name becomes ambiguous across nested siblings.
> 
> **Overview**
> Enables **nested processes to share a background `action_name`**, routed at phase 1 by conditions (same polymorphism as sync). Phase 1 now persists **`owning_process_class`** on `TransitionMessage` (migration `0007`); phase 2 **`_find_transition`** restores only **`is_background`** transitions from that owner instead of first-match by name.
> 
> **Validator** `_validate_unique_background_action_names` now rejects only **two background transitions with the same name in one process class**; duplicates across distinct nested classes and **sync/background name sharing** are allowed (phase 1 still disambiguates ambiguous calls with `TransitionNotAllowed`).
> 
> **Legacy / owner-less rows** fall back to name lookup only when the name is **unique** in the tree; if ambiguous, restore **refuses to guess** (`_RestoreError` → row finalized, no side-effects). **`serialize_kwargs`** and snapshots keep `owning_process_class` off side-effect kwargs.
> 
> Docs (README/CHANGELOG) and broad tests cover Gmail/Dummy routing, Celery boundary, watchdog, and concurrency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80579056ae16dfc43b05eaadb8e2d10d1c48b434. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


emil-balashov:
## Go-live hardening (second review pass) — `881b7b1`

A deeper, production-readiness adversarial review (4 dimensions × 3 verifiers each, focused on Celery-mode durability, multi-worker concurrency, migration safety, restore edge cases) surfaced **6 confirmed findings beyond the first pass's test-only gaps**. All are now fixed:

### Correctness / durability
- **Owner-less restore no longer guesses (findings #1, #2, #5 — one `high`).** When a `TransitionMessage` has no resolvable `owning_process_class` — a legacy/pre-discriminator row, or one whose owning nested process was renamed/removed mid-flight — `runner._find_transition` previously fell back to first-match. With the relaxed validator now allowing a shared background `action_name` across nested processes, that could run the **wrong** integration's side-effects (a `BackgroundAction`, where the state guard can't tell siblings apart by `sources`) or strand the instance. It now resolves by name **only when the name is unambiguous across the tree**; if ambiguous it raises `_RestoreError`, finalizing the row without side-effects (contained, retries stopped). Unique-name legacy rows are unaffected.
- **`owning_process_class` is now `TextField` (was `CharField(255)`) (findings #4, #6).** A deeply-namespaced dotted path could exceed 255 and raise `DataError` in phase 1 — not an `IntegrityError`, so it escaped the `AlreadyInProgress` guard and surfaced as an unhandled 500, never enqueuing the transition. `TextField` matches the unbounded `process_class` and is never indexed.

### Deploy safety / docs
- **Migration `0007` lock note (finding #3).** Documented the brief `ACCESS EXCLUSIVE` lock on the hot `transitionmessage` table (additive, non-rewriting on PG 11+) and the `lock_timeout` recommendation under load — in the migration docstring + CHANGELOG/README.
- **Upgrade note:** drain in-flight rows before refactoring a unique-name background transition into the shared-name nested pattern.
- Corrected an inaccurate comment (the owner is recorded for *every* background transition, not blank for bound-process ones).

### Production-fidelity validation
There is no deployable Heroku app for this library (CLAUDE.md's Heroku validation was run via a downstream app); I validated against the equivalent stack the durability contract is defined against — **real PostgreSQL 17 + Redis 7.2**:
- Full suite on PostgreSQL + Redis: **356 tests** ✅ (and 356 on SQLite)
- **Stability suite** (real threads, `select_for_update(nowait)`, partial-unique guard, worker-crash/deadlock injection): **58 tests** ✅
- New coverage: **Celery cross-process round-trip** (phase 1 records owner + enqueues; a fresh worker invocation restores from the committed row alone), **Postgres real-thread concurrency** (no owner bleed; one-in-flight guard holds), **safety-net** finalizes the correct sibling's `failed_state`, and **owner-less ambiguity containment** (runs neither sibling; unique-name still resolves).
- `makemigrations --check` clean; migration `0007` applies on a fresh DB.

Issue #125, opened 2026-07-17
Bindings registry + Django system checks: make bind-time hook warnings impossible to miss


emil-balashov:
Two related gaps found while adopting 0.5.1 in gv:

1. **Warn-mode hook validation is invisible exactly when it matters.** `_validate_hook_signatures` warns through the transition logger at `bind_model_process` time — which runs during `AppConfig.ready()`, before test/dev logging is fully configured. gv's warnings-mode suite showed ZERO warnings while strict mode found three real offenders on the same tree. A logger nobody sees at bootstrap is not a guardrail.

2. **There is no first-class way to enumerate bound machines.** gv's coverage audit had to introspect model class properties and their closures to reconstruct what was bound. Consumer tooling (coverage matrices, contract tests, admin) keeps needing this.

## Proposal
- `ProcessManager.bindings` — a public registry of `(model, process_class, state_field)` appended by every `bind_model_process` call.
- `django_logic/checks.py` — a Django system check (tag `django_logic`, id `django_logic.W001`) that walks the registry and re-emits every hook-signature offender through the checks framework: surfaced by `manage.py check`, **every test run**, and deploy checks — regardless of logging config. Strict mode keeps raising at bind; the check covers warn-mode consumers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #126, opened 2026-07-17
Bindings registry + django_logic.W001 system check (#125)

emil-balashov:
Closes #125.

- **`ProcessManager.bindings`**: public registry of `ModelProcessBinding(model, process_class, state_field)`, appended by every `bind_model_process`. Consumer tooling (gv's coverage audit had to reconstruct bindings from model-property closures) gets a first-class surface.
- **`django_logic.W001`**: a system check (tag `django_logic`) that re-runs `collect_hook_signature_offenders` over the registry and reports through Django's checks framework — visible in `manage.py check`, **every test run**, and deploy checks, independent of logging config. Motivation: bind-time warnings are logged during `ready()` before logging is configured; gv's warn-mode suite showed **zero** warnings on a tree where strict mode found three real offenders.
- Bind-time behavior unchanged (warn by default, `STRICT_HOOK_SIGNATURES` raises); the validator's collection half is now reusable (`collect_hook_signature_offenders`).

Suite: 358 tests OK. Binding tests now also clean the registry (it's global state).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Additive registry and read-only system check; bind-time validation behavior is preserved with extracted collection logic.
> 
> **Overview**
> Adds a **first-class bindings registry** and a **Django system check** so hook-signature problems are visible even when bind-time logger warnings are lost during `AppConfig.ready()`.
> 
> Every `ProcessManager.bind_model_process` call now appends a **`ModelProcessBinding`** `(model, process_class, state_field)` to **`ProcessManager.bindings`**, giving consumer tooling a stable list instead of inferring bindings from model properties.
> 
> Hook validation is refactored: **`collect_hook_signature_offenders`** owns the inspection logic; bind-time **`_validate_hook_signatures`** still warns or raises under **`STRICT_HOOK_SIGNATURES`** unchanged. **`django_logic.W001`** re-runs that collection over the registry via **`django_logic/checks.py`**, registered when the background app's **`ready()`** imports checks — surfaced by **`manage.py check`**, test runs, and deploy checks.
> 
> Tests clear **`ProcessManager.bindings`** in teardown and add **`tests/test_system_checks.py`** for registry and W001 behavior.
> 
[...]

PR #133, opened 2026-07-17
Transition-execution coverage: observers + coverage report (#132)

### LINKED ISSUE #132: Transition-execution coverage: record what a test suite actually drives, report what it never touched
## Problem

Consumers auditing FSM test coverage today must resort to static analysis (grepping test trees for `process.action()` calls / ProcessScenario drives). That approach has two blind spots, both hit in practice during the gv #8577 migration audit:

1. **Transitive execution is invisible** — a test that drives a view/task/handler which calls `instance.process.action()` counts as *uncovered* statically. In gv, ~150 of 167 statically-uncovered pairs turned out to be executed at runtime.
2. **Dynamic dispatch is invisible** — `getattr(process, action_name)()` (batch actions, import-sheet routing) can't be attributed statically at all.

The engine is the single choke point that knows ground truth: every initiation passes through `Process._get_transition_method`, which already resolves `(transition, owning_process)` — including the exact nested process class for condition-disambiguated trees.

## Proposal

1. **Core: transition-initiation observers.** A module-level `transition_observers` list in `django_logic.process`; `_get_transition_method` notifies each observer with `(owning_process_cls, action_name, instance)` after successful resolution. Zero overhead when empty; a raising observer is logged and never breaks the transition. (Generally useful beyond coverage: metrics, tracing.)
2. **`django_logic/coverage.py`** (top-level, no test-framework imports so it is safe from `AppConfig.ready()`):
   - `iter_bound_transitions()` — walk `ProcessManager.bindings` incl. nested processes, yielding every declared transition with its owning class.
   - `TransitionCoverage` — in-memory context-manager recorder for single-process runs.
   - File-backed recording activated by `DJANGO_LOGIC['TRANSITION_COVERAGE_LOG'] = <path>` (activated in the existing `BackgroundConfig.ready()` beside the checks import) — append-only unique pairs per worker, works under both fork and spawn parallel test runners.
   - `coverage_report(executed=None, log_path=None)` — diff against the bindings registry: totals, executed, uncovered (with background flag), so a consumer CI job can fail when a previously-driven transition goes silent.

## Notes

- Initiation semantics: phase-2 background restore does not re-notify (phase 1 already recorded the pair). `retry`/rerun paths therefore don't inflate counts.
- Born from the gv migration audit where this exact hook (prototyped by wrapping `_resolve_transition_with_owner`) produced the runtime ground-truth list the static audit couldn't.
- Follows #125/#126: the bindings registry was added precisely so tools like this don't re-derive bindings from model internals.

emil-balashov:
Closes #132.

## What

1. **Core: transition-initiation observers.** `django_logic.process.transition_observers` — a public list of callables invoked as `observer(owning_process_cls, action_name, instance)` after `_get_transition_method` resolves a transition, before it executes. Covers every initiation path: direct calls, `next_transition` follow-ups (they re-enter through the Process entrypoint), background phase 1. Phase-2 restore/retries do **not** re-notify — phase 1 already recorded the initiation. Zero overhead when the list is empty; a raising observer is logged and never breaks a transition.

2. **`django_logic/coverage.py`** (top-level module, no test-framework imports — safe to import from `AppConfig.ready`):
   - `iter_bound_transitions()` — walks `ProcessManager.bindings` (#125) including nested processes, yielding every declared transition with its owning class.
   - `TransitionCoverage` — in-memory context-manager recorder for single-process runs.
   - File-backed recording via `DJANGO_LOGIC['TRANSITION_COVERAGE_LOG'] = <path>`, activated in the existing `BackgroundConfig.ready()` beside the checks import — spawn-based parallel workers self-activate, fork-based workers inherit the parent's recorder; workers append unique pairs, duplicates across workers are merged at report time.
   - `coverage_report(executed=None, log_path=None)` — totals + sorted `uncovered` entries (`process`, `action`, `background`, `models`), ready for a CI gate that fails when a previously-driven transition goes silent.

## Why

Consumers auditing FSM coverage statically hit two blind spots (both found during the gv #8577 migration audit): transitive execution (test → view/task → `process.action()`) counts as uncovered, and dynamic dispatch (`getattr(process, name)()`) can't be attributed at all. Prototyping this exact hook against the gv suite showed ~150 of 167 statically-"uncovered" pairs actually execute at runtime — and produced the exact short list that doesn't.

## Tests

7 new tests in `tests/test_coverage.py`: observer payload + **nested owner attribution** (the declaring process class, not the bound root), raising-observer isolation, nested walk of declared transitions, executed/uncovered split, file-backed dedup + merge, idempotent activation. Full suite: **373 OK** (was 366).

---

> [!NOTE]
> **Low Risk**
> Observers run on every initiation when registered but are isolated from transition success; file recording is opt-in via settings.
> 
> **Overview**
> Adds **runtime transition-execution coverage** so test suites can see which declared FSM transitions were actually initiated, including transitive and dynamic drives that static analysis misses.
> 
> The engine now exposes **`transition_observers`** and fires **`(owning_process_cls, action_name, instance)`** after a transition resolves in **`_get_transition_method`** (direct calls, **`next_transition`**, background phase 1 only). Raising observers are logged and do not block the transition; there is no work when the list is empty.
> 
> **`django_logic.coverage`** compares recorded **`(process, action)`** pairs to all transitions from **`ProcessManager.bindings`** (nested processes included): **`TransitionCoverage`** for in-memory runs, optional **`DJANGO_LOGIC['TRANSITION_COVERAGE_LOG']`** with **`coverage_report()`** for parallel tests, activated idempotently from **`AppConfig.ready()`** on both **`django_logic`** and **`django_logic.background`**.
> 
> Docs and changelog describe CI usage and the append-only log footgun.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9b019c7f3ae870a0ba20a782f479d61746c21c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Transition-execution coverage** (#132). The resolver now notifies
+  `django_logic.process.transition_observers` with
+  `(owning_process_cls, action_name, instance)` on every transition
+  initiation (direct calls, `next_transition` follow-ups, background
+  phase 1; phase-2 restore does not re-notify). A raising observer is
+  logged and never breaks the transition. On top of it,
+  `django_logic.coverage` records executed `(process, action)` pairs and
+  diffs them against every transition declared by `ProcessManager.bindings`
+  (nested processes included): `TransitionCoverage` (in-memory context
+  manager), `DJANGO_LOGIC['TRANSITION_COVERAGE_LOG'] = path` (file-backed,
+  fork/spawn-safe parallel test runs), and `coverage_report()`. Static
+  test-tree analysis cannot see transitive or dynamically-dispatched
+  drives; the engine can — this answers "which transitions did the suite
+  never drive?" exactly.
+
 ### Fixed
 
 - **NextTransition no longer forwards `request` into background
--- a/README.md
+++ b/README.md
@@ -681,6 +681,7 @@ DJANGO_LOGIC = {
     'SENTRY_TRANSACTION_NAMING': True,  # per-transition Sentry naming (no-op without sentry-sdk)
     'STRICT_KWARGS_SERIALIZATION': False,  # True: raise (not warn) on dropped 'request' / non-string dict keys
     'STRICT_HOOK_SIGNATURES': False,    # True: refuse to bind hooks without a named instance-first parameter
+    # 'TRANSITION_COVERAGE_LOG': '...',  # opt-in: record driven transitions to a file (see "Transition-execution coverage")
 }
 ```
 
@@ -1195,6 +1196,10 @@ Diffing `report['uncovered']` in CI catches transitions that silently stop
 being exercised. The observer list is public — consumers can register their
 own hooks (metrics, tracing); a raising observer is logged and never breaks a
 transition.
+
+The log is **append-only and never truncated** — point each run at a fresh
+path (or delete the old file first), or stale pairs from earlier runs count
+as covered.
 
 ## Contributing
 Pull requests are welcome. For major changes, please open an issue first to discuss what you would like to change.
--- a/README.md
+++ b/README.md
@@ -1157,6 +1157,45 @@ class TestStuckOrder(ProcessScenario):
 
 **The full guide — [docs/TESTING_GUIDE.md](docs/TESTING_GUIDE.md)** — documents every test scenario for a process (happy paths, gating, failures, retries, terminal failures, one-in-flight conflicts, superseded rows, nested processes, snapshot replay) with copy-pasteable examples, and explains the philosophy: **you test your process; the library guarantees the background machinery** (validated by its own regression suite and a production-style Heroku matrix), so your tests never need a Celery broker.
 
+### Transition-execution coverage
+
+Which transitions does your test suite actually drive? Static analysis of the
+test tree can't tell — a test that exercises a view or Celery task which calls
+`instance.process.action()` looks uncovered, and dynamically-dispatched drives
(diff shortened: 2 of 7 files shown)


emil-balashov:
Reviewed with a three-lens fleet (hot path / coverage module / API+tests) plus adversarial verification, then fixed the confirmed findings in `89834fc`:

**Engine hook: verified correct.** All three lenses independently traced and probe-tested the exactly-once contract — the single notify site in `_get_transition_method` covers direct calls, `next_transition` follow-ups (re-enter via the Process entrypoint), and background phase 1; phase 2 (`_restore`/`_find_transition` → direct hook execution) structurally cannot re-notify. Zero overhead when the list is empty; observer exceptions isolated before any lock is taken.

**Fixed before merge:**
- **Silent no-op without the background app** (confirmed major, flagged by all three lenses): checks registration + coverage activation now also live in a new top-level `django_logic` AppConfig — this also closes the W001 gap noted on #126.
- **Headline semantics were untested**: added pins for background exactly-once, no re-notify across MAX_ERRORS retries, follow-up self-attribution, dirty-exit cleanup, raising-observer isolation.
- `coverage_report(log_path=...)` no longer crashes on a never-written log; `_FileRecorder` marks pairs seen only after a successful append.
- Documented: append-only log needs a fresh path per run; same-class sync/bg namesakes collapse to one key; `TRANSITION_COVERAGE_LOG` added to the README settings block.

**Accepted as designed:** initiation-time recording counts lock-refused initiations as driven (documented semantic — the module docstring now says so explicitly).

Combined with master (0.7.0): 402 tests OK.

Issue #146, opened 2026-07-22
Give transition coverage a stable identity beyond action_name


emil-balashov:
## Problem

Transition execution coverage keys declarations and observations only by `(owning process class, action_name)` in `django_logic/coverage.py`.

The engine legally supports condition-disambiguated same-name transitions, including a synchronous and background transition sharing one name in a class. Those declarations collapse into one coverage entry. Driving one marks all namesakes covered, undercounts `total`, and can report the wrong `background` value.

The module docstring currently documents this as a footgun rather than preserving declaration identity.

Related: #132 introduced transition coverage, but did not cover this legal topology.

## Acceptance criteria

- Assign each declared transition a stable identity that distinguishes legal same-name declarations.
- Observer events carry enough information to identify the selected transition, not only its action name.
- Reports count and classify every declaration separately.
- Tests cover condition-disambiguated same-name transitions and sync/background namesake pairs.
- Define compatibility for existing file-backed coverage logs.

emil-balashov:
Implemented on the PR #137 branch in 9ba9559: coverage keys now carry the declaration's kind and shape (class + action + sync/bg + sorted sources → target), so condition-disambiguated namesakes — including a sync+background pair in one class — count and cover separately; report entries expose sources/target. Observers receive the resolved transition as a fourth argument (breaking for the 0.8 observer API — CHANGELOG upgrade note; both built-in recorders updated). Compatibility for existing file logs: 2-field 0.8 lines keep their original cover-all-namesakes semantics, so old logs don't churn to uncovered. Closes with #137.

PR #153, opened 2026-07-23
Bugbot round 4: contained unlock cleanup, condition-aware coverage keys, per-state warn-once

emil-balashov:
Follow-up to #137, which merged while Bugbot's final pass was in flight — its three findings (all on #137's last commits, all confirmed real) land here instead.

## Fixes

1. **Savepoint unlock cleanup not exception-safe** (Medium, `commands.py`) — `_run_in_savepoint`'s rollback cleanup released the discarded deferred unlocks in a bare loop: one release raising (a cache/Redis blip) skipped the remaining sibling releases *and replaced the hook's original exception*. Each release is now contained — logged, degrading to the documented TTL-bounded leak — and the original exception always propagates. Pinned by `test_savepoint_cleanup_survives_a_failing_unlock` (first unlock blips → second still released, hook exception preserved).

2. **Coverage collapses same-shape namesakes** (Medium, `coverage.py`) — declaration keys carried kind + `sources→target` but not conditions, so same-class namesakes differing **only** by conditions still collapsed — and that is exactly the per-courier polymorphism pattern (gv's `dispatch ×N` variants share shape and differ only by courier conditions), so the #146 promise didn't hold where it mattered most. Keys now include the condition callables' sorted qualnames (anonymous lambdas can still collide; named condition functions — the norm — stay distinct). Pinned by `ConditionOnlyNamesakeIdentityTests` (two condition-only variants count separately; driving one covers only it).

3. **Namesake missed failed_state warn** (Medium, `dispatch.py`) — the warn-once key was `(model, field, action_name)`, so namesake transitions parking candidates in *different* `in_progress_state`s shared it: only the first backlog ever surfaced. `in_progress_state` joins the key; each parked backlog warns with its own candidate count. Pinned by `test_namesakes_with_distinct_in_progress_states_each_warn`.

## Validation

- Engine suite: **537 tests OK** (sqlite; +4 from the new regressions).
- django-logic-test harness (editable sibling install of this branch): **75 passed** — TC-26/27/28 unaffected.
- CHANGELOG entries added under `[Unreleased]`.

The three Bugbot threads on #137 will be answered with pointers here.

---

> [!NOTE]
> **Low Risk**
> Localized fixes to hook savepoint cleanup, coverage key/report matching, and logging keys for stranded recovery; behavior changes are edge-case hardening with regression tests and backward-compatible coverage log handling.
> 
> **Overview**
> **Three Bugbot follow-ups** harden savepoint rollback, transition coverage identity, and stranded-state observability.
> 
> **Savepoint unlock cleanup** (`_run_in_savepoint`): when a hook savepoint rolls back and releases deferred `DEFER_UNLOCK_UNTIL_COMMIT` locks, each `unlock()` is now isolated in its own try/except. A cache blip on one release no longer skips sibling releases or replaces the hook’s original exception; failures are logged and degrade to the documented TTL-bounded leak.
> 
> **Coverage declaration keys** now append a **conditions fingerprint** (sorted condition callables’ qualnames) so same-class namesakes that share `sources→target` but differ only by `conditions` count and cover separately. **`coverage_report`** still accepts older log lines: 0.8 two-field keys cover all same-name namesakes; 0.9.0 four-field keys (no fingerprint) cover every declaration sharing that prefix so upgraded logs are not falsely all-uncovered.
> 
> **Stranded recovery warnings** for transitions without `failed_state` are warn-once per `(model, state_field, action_name, in_progress_state)` instead of per action name, so namesake transitions parking in different in-progress states each surface their backlog once.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 246a3e6a503d51331c5f49eb5d44e74c5b3fe387. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Savepoint unlock cleanup is exception-contained.** When a hook
+  savepoint rollback releases the deferred unlocks it discarded (#141 ×
+  #138), one release raising (a cache blip) no longer skips the sibling
+  releases or replaces the hook's original exception — the missed
+  release degrades to the documented TTL-bounded leak, logged.
+
+- **Coverage keys carry a conditions fingerprint** (#146 follow-up).
+  Same-class namesakes sharing sources→target and differing only by
+  `conditions` — the per-courier polymorphism pattern — no longer
+  collapse into one declaration: the key includes the condition
+  callables' sorted qualnames. (Anonymous lambdas can still collide;
+  named condition functions, the norm, stay distinct.)
+
+- **The missing-`failed_state` warn-once key includes
+  `in_progress_state`** (#145 follow-up). Namesake transitions parking
+  candidates in different states are different parked backlogs — the
+  second is no longer silenced by the first's warning.
+
 ## [0.9.0] — 2026-07-23
 
 Stranded-state recovery (#136): `recover_stranded_states`, the fifth
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,12 @@
   `conditions` — the per-courier polymorphism pattern — no longer
   collapse into one declaration: the key includes the condition
   callables' sorted qualnames. (Anonymous lambdas can still collide;
-  named condition functions, the norm, stay distinct.)
+  named condition functions, the norm, stay distinct.) The wider key
+  stays backward-compatible with persisted coverage logs: a 0.9.0-format
+  line (no conditions field) covers every declaration sharing its
+  `class⇥action⇥kind⇥sources>target` prefix, so a log carried across the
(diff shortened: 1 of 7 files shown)


emil-balashov:
Bugbot's coverage-key finding (0.9.0-format 3-tab log lines no longer matching the new 4-tab declaration keys → false 'uncovered' gaps after upgrade) is addressed in bd69aa3: `coverage_report` now rescues 0.9.0-format lines the same way it already rescues 0.8 legacy lines (a 0.9.0 line covers every declaration sharing its 4-field `class⇥action⇥kind⇥sources>target` prefix). Pinned by two regressions in `ConditionOnlyNamesakeIdentityTests`/`NamesakeDeclarationIdentityTests` that both fail without the shim. Verified locally: 539 tests pass.

Issue #151, opened 2026-07-23
Make RedisState lock refresh fully atomic under TTL overrun


emil-balashov:
## Context

Came out of review of PR #137 / #139 lock ownership tokens.

`RedisState.set_state` xx-refreshes the single value+lock key. A non-atomic read-merge-write of the ownership token can, under TTL overrun, plant a stale holder's token over a successor's lock so a late `unlock()` CAD-deletes the live lock (the dual-entry hazard #139 targets).

## Current mitigation (PR #137 follow-up)

Token-gated refresh: if this `State` object holds `_lock_token` and the stored token is a different value, skip the cache rewrite. That closes the common **post-takeover** `set_state` → wrong-token → late `unlock` path and is covered by unit tests.

Docs/CHANGELOG no longer claim "never a wrong unlock."

## Residual gap

The get → compare → set sequence is still not multi-process atomic. A takeover strictly between the ownership check and the write can still rewrite a successor's token.

## Proposed work

Pick one (or equivalent):

1. **Redis Lua / `EVAL`** via django-redis that updates state + token only when the stored token still matches the expected token (or key missing → no-op for xx semantics).
2. **Split keys**: token key (CAD target) separate from state value key so `set_state` never rewrites ownership.
3. Document as permanent best-effort if we deliberately accept TTL-overrun races as out of contract (pair with guidance: size `LOCK_TIMEOUT` / per-transition `lock_timeout` above max side-effect time).

## Acceptance

- Regression test for post-takeover `set_state` stays green (already exists).
- If Lua/split-keys: a test or stress note that a concurrent takeover during refresh cannot leave the stale token on the successor's key.
- CHANGELOG/contract text matches the actual guarantee.

## Related

- #139 lock ownership tokens
- PR #137 stranded recovery / hardening batch
- Bugbot: "Token race allows wrong unlock" on `django_logic/state.py`

emil-balashov:
Addressing this via **option 1 (Redis Lua compare-and-set)**, not split-keys — split-keys would break the single-key immediate-visibility contract and add two-key consistency skew (state key lingering after the token key expires).

[...]

Issue #139, opened 2026-07-22
Use ownership tokens so expired lock holders cannot unlock successors


emil-balashov:
## Problem

State locks do not record ownership:

- `State.lock()` stores `True`; `unlock()` unconditionally deletes the cache key.
- `RedisState.lock()` stores the state value, but `unlock()` is still an unconditional delete.

If transition T1 exceeds its TTL, T2 can acquire the expired lock. When T1 eventually finishes, its `unlock()` deletes T2's lock. T3 can then enter concurrently with T2.

This is distinct from the pre-lock TOCTOU fixed by #86 and lock cleanup fixed by #85: the stale owner did acquire a valid lock, but no longer owns the successor lease it deletes.

Affected code: `django_logic/state.py`.

## Impact

Mutual exclusion can collapse precisely during slow work or pauses—the same TTL boundary used by stranded-state recovery in PR #137.

## Suggested direction

Use a unique owner token per acquisition and atomic compare-and-delete on unlock. Consider fencing or lease renewal for work that can exceed its expected duration.

## Acceptance criteria

- Every successful acquisition has a unique ownership token.
- Unlock deletes only the token owned by that `State` instance.
- A stale owner cannot delete a successor's lock after TTL expiry/takeover.
- Tests cover T1 expiry, T2 takeover, T1 late unlock, and T3 exclusion for both supported lock implementations.

emil-balashov:
Implemented on the PR #137 branch in e77b654: every acquisition stores a unique ownership token and unlock() is a compare-and-delete — T1-expiry → T2-takeover → T1-late-unlock now leaves T2's lock intact (T3 stays excluded). RedisState wraps its single value+lock key as {state, token}; set_state preserves the stored token on holder and non-holder writes (a race there degrades to a TTL-bounded leak, never a wrong unlock); raw 0.8 values stay readable — deploy web+workers together. Tokenless State objects keep unconditional delete as the manual force-release path. Tests cover expiry/takeover/late-unlock/exclusion for both implementations (tests/test_lock_ownership.py), validated against real django-redis in the settings_redis suite. Closes with #137.

Issue #81, opened 2026-06-05
First-class recipe (and maybe helpers) for parent/child coordination without nested process calls


emil-balashov:
The original pain (`fundamental problem.md`) is nested `other.process.xxx()` calls inside side-effects cascading failures across state machines. We validated a clean alternative for an order with many fulfillments:

1. parent **fans out** — starts each child's own `BackgroundTransition` (phase 1) and returns; no child work runs in the parent's transition.
2. child failure is **contained** in the child's own `failed_state` (+ error on the row).
3. children **report back via best-effort callbacks** that run an **idempotent, guarded completion check** on the parent.
4. parent **aggregates errors by reading child rows**, and has an explicit `action_required` partial-failure state. No child exception is ever raised into the parent.

Result: all children succeed → order `fulfilled`; one fails → order `action_required` naming the failed child, siblings unaffected.

**Proposal:** make this a first-class section in the docs, and consider a small helper (e.g. a "when all children terminal, fire transition X" completion utility). Full write-up + tests: [django-logic-test `NESTED_PROCESS_ERROR_HANDLING.md`](https://github.com/Borderless360/django-logic-test/blob/main/docs/design/NESTED_PROCESS_ERROR_HANDLING.md).

See `docs/IMPROVEMENTS_FROM_HEROKU_VALIDATION.md` (#4).

emil-balashov:
Merged into `release/0.3.0` via #82 (auto-close keywords only fire on the default branch). Validated on Heroku — see RESULTS in django-logic-test.

PR #172, opened 2026-07-27
0.10.0 — overengineering sweep: −4.5k lines, 4 settings retired, W002 added

### LINKED ISSUE #157: Diet: delete 5 stale planning/research docs + docs/INDEX.md (~2,311 lines)
Delete five pre-implementation/research docs plus the index whose only job is tiering them into "current" vs "historical".

**Files:** `docs/PLAN.md`, `docs/research/PR-75-REVIEW.md`, `docs/research/idea1.txt`, `docs/design/TESTING_SCENARIOS.md`, `docs/IMPROVEMENTS_FROM_HEROKU_VALIDATION.md`, `docs/INDEX.md` — **~2,311 lines**.

**Why safe:** not packaged (no `MANIFEST.in`; `SOURCES.txt` contains no `docs/`), no test reads them, and consumers vendor no `.md`.

**Why now — the rot is already realized, not hypothetical:**
- `docs/PLAN.md` §2.10's settings block omits **five shipped knobs**, while `CLAUDE.md:96` instructs anyone changing the engine to read it. That is an active correctness hazard.
- `TESTING_SCENARIOS.md` is linked from the README as "Full design" but names functions that do not exist and documents 8 of 18 assertions.
- Docs total 6,033 lines against 6,852 engine LOC.

**Sequencing — do these or the deletion leaves dangling links:**
- Repoint to `docs/TESTING_GUIDE.md`: `README.md:1168`, `django_logic/testing/__init__.py:22`
- `CLAUDE.md:96` → point at `docs/design/BACKGROUND_TRANSITION_ANALYSIS.md` only; also `CLAUDE.md:100-101`
- `docs/design/BACKGROUND_TRANSITION_ANALYSIS.md:4`
- Before deleting `IMPROVEMENTS_FROM_HEROKU_VALIDATION.md`, move its two still-open items into `TODO.md`: ERROR→WARNING triage for handled safety-net paths, and the Postgres connection-budget note.
- **Keep** `docs/research/race-condition-issue` — `tests/stability/test_deadlocks.py:13` and `test_race_conditions.py:14,130` cite it as their reproduction source.

Nothing unique is lost from `PLAN.md`: `CHANGELOG.md:480` carries the mixed-version deploy caveat verbatim, and README+CHANGELOG carry `STRICT_KWARGS_SERIALIZATION` and `__dl_type__`.

### LINKED ISSUE #158: Diet: delete BaseTransition and the 4 *_class hooks with no override
`BaseTransition` (`django_logic/transition.py:47-79`) contributes nothing but a name.

- Four methods that only `raise NotImplementedError`, all four overridden by `Transition`.
- Seven `*_class` swap hooks, four of which have **zero overrides anywhere**.
- `is_background`.

**Zero references** in the library, its test suite, or either consumer tree. `BackgroundTransition` extends `Transition`, not `BaseTransition`. It is not in `__init__.py`'s re-export list and no consumer imports it, so the CHANGELOG's earlier deferral of this merge as "breaking" no longer holds.

**Origin:** inherited from the pre-0.3 design when `Transition` had `background_mode`/`run_in_background` siblings (see `TODO.md` 0.3.0: "Remove `background_mode` / `run_in_background` from base `Transition`"). The last sibling implementation went away; the base did not.

**Do:**
1. Delete `class BaseTransition`; move `is_background = False` and the kept `*_class` attributes onto `Transition` directly — they must stay as class attributes because `Transition.__init__` reads them.
2. Same pass: drop the four `*_class` hooks with no override anywhere — `conditions_class`, `failure_callbacks_class`, `failure_side_effects_class`, `next_transition_class` — and instantiate those bundles directly.
3. **Keep `Process.permissions_class`** — consumers subclass it (`StrictPermissions` / `NoUserPermissions`, 5 live sites).

~33 LOC. **Verified:** a reviewer executed this change; suite 540 OK and the result was byte-identical. All 8 `is_background` reads go through `getattr`, so nothing breaks.

Non-breaking for both consumers, but it is a non-underscore name on a published PyPI package — deserves a CHANGELOG "Removed" line.

### LINKED ISSUE #159: Diet: delete the duplicated failure-side-effect savepoint layer in runner.py
`runner._FailureSideEffectsRollback` + `_run_failure_side_effects_isolated` (`django_logic/background/runner.py:663-695`) duplicate a rule that #138 moved into `commands.py` with a **stronger** savepoint — the `commands.py` version also releases deferred unlocks that a rollback discards.

Both call sites are already inside `atomic()`, so the older local layer adds nothing.

~33 LOC. **Verified:** a reviewer executed the deletion — suite 540 OK — and separately confirmed the surviving layer is non-vacuous (it does real work, so this is not "delete both").

### LINKED ISSUE #160: Diet: delete coverage.py's 0.8 + 0.9.0 cross-version log readers
`coverage.py` carries readers for two older log key formats: `legacy_pairs` (0.8 two-field) and `prior_full` (0.9.0 four-field) — `django_logic/coverage.py:47-53,134-148,159,167-168`.

**These guard a one-day window between tags.** 0.9.0 and 0.9.1 were released a day apart, the documented practice is a fresh log per run, and nobody has ever persisted a coverage log across a version bump. The report schema is unchanged by removing them.

~70 LOC and one concept ("cross-version key compatibility") that a reader currently has to learn.

**Verified:** a reviewer deleted these and only the three tests that exist to test the readers themselves failed — no other collateral.

Part of a wider pattern (see the compatibility-layer sweep): four independent compat layers exist for versions no consumer has ever run.

### LINKED ISSUE #161: Diet: delete the non-finite float pre-scan in background/serializers.py
`django_logic/background/serializers.py:169-186,247-255` walks the kwargs tree to find non-finite floats (`_non_finite_float_paths` / the `bad_floats` block) so it can name the offending key path in the error.

Two lines below, `json.dumps(allow_nan=False)` rejects a **superset** of the same inputs — including non-finite values in dict *keys*, which the manual scan cannot reach — and already raises the same `TypeError`.

~45 LOC.

**Verified:** a reviewer deleted the pre-scan and only four assertions on the error-message text failed. The single functional loss is the offending key path in the message; CPython still names the offending value.

### LINKED ISSUE #162: Diet: de-duplicate prose blocks README/CHANGELOG already own (corrected scope)
De-duplicate the prose blocks that `README.md` / `CHANGELOG.md` already own, and trim the duplicated README testing chapter.

**Scope (this is the corrected scope — see below):**
- `transition.py:298-329` — `_release_lock`'s docstring repeats the `DEFER_UNLOCK_UNTIL_COMMIT` trade-off list verbatim from `README:931-935` and `CHANGELOG:103-114`, and the docstring itself says "(see README)". ~20-25 lines.
- `dispatch.py:98-102,142-187,362-391,412-414`; `runner.py:35-41,150,852-876`; `state.py:246-256`; triplicated rationale across `conf.py`/`apps.py`.
- `README.md:843-861` and `1037-1171` — duplicated testing chapter; `docs/TESTING_GUIDE.md` is the live superset. **Keep the `sync_execution()` snippet** (consumers call it in production).
- Normalise the two pre-fix-narration lines: `transition.py:303`, `process.py:112`.

~300 lines total (~170 in-code + ~130 README).

**Read this before widening the scope.** Five reviewers independently proposed a much larger cut ("~400 lines of issue archaeology"); verification measured that thesis as **~50x overstated**: across process.py + transition.py's ~490 prose lines, only **8** carry an issue number and 5 are already the bare `(#NNN)` form. Prose density is a uniform house style (`coverage.py` 50%, `dispatch.py` 47%, `state.py` 42%, `transition.py` 42%), not damage concentrated by the hardening rounds. Several of the fattest docstrings exist specifically to stop a future simplifier flipping `is True` back to truthiness or moving the TTL memo into `lock()`'s failure path — each with a named pinning test. **Do not apply a density target.**

Docstrings here are functionally inert (no sphinx/autodoc, no doctests, zero `__doc__` reads in any tree), so this is safe.

Related: three reviewers cited a `CLAUDE.md` rule forbidding narrating comments. **That rule does not exist in this repo** (`grep -rn narrat` over all `.md`/`.mdc` → zero matches). Either add it to `CLAUDE.md` or stop invoking it.

### LINKED ISSUE #163: Diet: dead-code sweep — 8 small verified deletions
Eight small deletions, each independently verified. Grouped because none justifies its own PR.

1. **`_infer_field_name`** (`background/runner.py:828-836`) — pre-0.4 row fallback. Instrumented: 1 hit across 540 tests, from its own self-test. An empty `field_name` now fails closed through the existing `_RestoreError` path. 39 LOC. *Verified: 540 → only its self-test fails.*
2. **`JourneyStep.matches()`** (`testing/scenario.py:66-74`) — hand-copy of the generated `@dataclass __eq__`; use `!=`. 64/64 truth-table agreement, no subclass anywhere. 9 LOC. *Verified: 540 OK.*
3. **`BackgroundAction.complete_transition`** no-op override (`background/transitions.py:286-292`) — unreachable: its only caller is `SideEffects.execute`, which background phase 1 never runs. Dead since d632da7. 7 LOC.
4. **`process_name` should be required** on the three testing message helpers (`testing/runner.py:49-91`). All six internal callers pass a `str`; not in `__all__`, not in README/GUIDE; `None` would crash the adjacent `getattr` anyway. The `None` default is documented as "historical unscoped behaviour" — a compat promise made in the same commit that updated every internal caller. It is worse than dead: it makes it possible to silently reintroduce the #150 cross-process bleed by omitting an argument. 15 LOC.
5. **Drop `ExecutionTracker` from `testing/__init__`** import + `__all__`, keep the class (`testing/__init__.py:27,35`). Zero importers in any tree; `track()` in `testing.tracking` is the supported path. 2 LOC.
6. **Delete `.cursor/rules/django-logic.mdc`** — zero audience (neither consumer has it) and five of its statements are now factually wrong (claims "no queue default", "four safety-net tasks" when 0.9.1 ships five). 128 LOC.
7. **Delete the 0.3.0-COMPLETE block from `TODO.md`** (keep the file and its INDEX pointer) — a parallel changelog of 15 released items that CHANGELOG already owns. 17 LOC.
8. **Delete `test_readme_has_no_stale_release_instructions`** (`tests/test_metadata.py:153-159`) and the `(0.9.0 at the time of writing)` parenthetical at `README.md:47`. Its sentinel `'legacy 0.1.x'` never existed in any README commit, so it is green while the drift it targets is live right now. **Keep** the Django-floor test. *Verified: 5 tests OK.*

### LINKED ISSUE #164: Diet: test/CI/packaging cleanup (settings dedupe, phantom stress step, broken dev extra)
Test, CI and packaging cleanup — no engine code involved.

1. **Test-settings dedupe.** 26 test modules each hand-copy the same 6-key `DJANGO_LOGIC` dict, plus 13 `@override_settings` that are no-ops. Replace with one `dl_settings(**overrides)` helper in `tests/__init__.py`. Safe because settings are read live on every call and there is no `setting_changed` receiver anywhere. **Verified: 540 OK, −209/+55.** ~154 LOC.
2. **Delete the re-declared `DJANGO_LOGIC` dict** in `tests/settings_redis.py:22-29` and `tests/settings_stability.py:41-48` — both already `import *` from `tests.settings`. **Verified:** namespaces diffed identical after removal. Also delete the unreferenced `stability-tests` service in `docker-compose.test.yml:26-40`, which is broken as written (sets a `DATABASE_URL` nobody reads). ~31 LOC.
3. **Delete the phantom nightly "stress – 10 threads" CI step** (`.github/workflows/stability.yml:76-90`). `STABILITY_STRESS_MODE` is read by nothing anywhere, so the step just re-runs the identical 61 tests. It has already produced a false claim in the harness `RESULTS.md:183`. ~15 LOC.
4. **Fix the `dev` extra** (`pyproject.toml:59,61,62`): drop `pytest`, `pytest-django`, `djangorestframework`. `pytest` is not merely unused, it is **broken** — a fresh venv gives "Interrupted: 48 errors during collection". **Keep the `drf` extra** (documented 0.4 migration path). Also fix the stale statements at `CLAUDE.md:91` and `CLAUDE.md:6-7`. *Verified: ran it.*

**Explicitly do not** add a `REDIS_URL` branch to `tests/settings.py`: `django_redis.clear()` is `flushdb()`, the ordinary unit suite calls `cache.clear()` in 10+ places, and this has already bitten us — the harness pops `REDIS_URL` before import for exactly this reason (`config/settings_test.py:14`).

Note for context: the test suite itself is **not** bloated. 13,449 test LOC against 6,852 engine LOC is a healthy 2:1 with 540 tests in 4.1s, and the two files that look like bloat (`test_issue_fixes_*.py`) are mutation-verified regression suites — 12 of 13 reverts are caught by the test naming the issue. Leave them alone.

### LINKED ISSUE #165: [1.0] Remove RedisState — 199 dead lines in state.py, and #139/#151 hardened code with zero users
**This is the single largest reduction available in the library, and it retires two hardening issues as dead code.**

`RedisState` is **199 of `state.py`'s 326 lines**, and no consumer has ever used it.

**Evidence:** neither consumer sets `state_class`. Measured on a live deployment carrying 28 bound machines / 137 transitions: **all 50 process classes resolve `state_class` to `State`**. There is no dynamic resolution path — a process either declares `state_class` or gets `State`.

**Delete:** `RedisState` (`state.py:128-326`), `get_db_state`, `_effective_timeout` / `_effective_lock_timeout`, `State._stored_token` (`state.py:24-28`) — the last three exist *only* to be read or overridden by the dead class. Plus its README section and ~380 test LOC / 33 tests. `state.py` 326 → 111.

**What this also collapses:**
- The three-reader footgun goes to two: `get_state` (cache-or-attr), `get_db_state` (cache-or-DB), `get_persisted_state` (DB-only).
- Explanatory comments in `dispatch.py`, `transition.py` and `testing/tracking.py` disappear.
- **#139** (token-in-key wrapper) and **#151** (the Lua compare-and-set — the top entry of the 0.9.1 release notes) turn out to be hardening of code with zero users. The base-`State` half of #139 is load-bearing and stays.

**No pending adoption path either:** the one consumer-side Redis state subclass is on the *legacy* engine fork, and its entire reason to exist (swap batch `update()` for a single `save()`) is already how the new `State` behaves.

**Verified:** a reviewer executed the deletion — 507 OK, zero collateral outside RedisState's own tests.

**Breaking — needs a major bump.** CHANGELOG entry:
> Removed: `RedisState`. It was an alternative cache-backed state backend; the default `State` has closed the stale-read gap it existed for (validate-then-lock TOCTOU). `get_db_state()` is also removed — use `get_persisted_state()`.

Ship `RedisState = State` for one release as cheap insurance.

**Two corrections to the naive how-to:**
1. **Keep `tests/settings_redis.py`** — its `requires_real_redis` gate also covers base-`State` tests that survive (`test_race_conditions.py:112/116/123`, `test_deadlocks.py:221`). Fix its now-false docstring at line 4.
2. **Do not** write `timeout or _get_lock_timeout()` — `or` swallows `timeout=0`. Keep `timeout if timeout is not None else ...`.
3. **Reword, don't delete,** `dispatch.py:423-425` and `state.py:33-35`: they are the only in-code markers of the b440162 fix, where `get_db_state` in the sweep read the sweep's own lock payload and would overwrite a legitimately advanced instance.

### LINKED ISSUE #166: [1.0] Remove 3 never-set settings: LOG_KWARGS*, PHASE2_STATE_GUARD='warn', SENTRY_TRANSACTION_NAMING
A settings census over the whole library: **16 keys are read, 6 are never set by any consumer** — and each of the 6 also carries a boot validator (#148/#149), README lines and dedicated tests. That is roughly a 4x tax on top of the knob itself.

This issue covers the three that are safe to delete outright. (`DEFER_UNLOCK_UNTIL_COMMIT` is the largest never-set knob and must **stay** — see the KEEP note at the bottom. `PROCESS_CLASS_ALIASES` is tracked separately because it is also half-implemented.)

**1. `LOG_KWARGS` + `LOG_KWARGS_REDACTOR`** (`logger.py:22-57`, `background/settings.py:211-235`) — delete both, plus `_validate_log_kwargs_redactor`, its tests, and the `docs/logger.md` section; `redact_log_kwargs` collapses to `dict(kwargs)`. Never set on any consumer branch, ever (`git log --all -S` finds only vendoring commits). **Keep the shallow copy** — that part is a real d0ee275 fix. ~160 LOC.
> CHANGELOG: Removed opt-in kwargs redaction. Consumers who had set `LOG_KWARGS_REDACTOR` will see un-redacted kwargs in log records.

**2. `PHASE2_STATE_GUARD='warn'`** (`runner.py:343-346,477-489` + `background/settings.py`) — the guard becomes unconditional. Its only purpose is restoring pre-0.4 behaviour, and the retired predecessor engine ran the guard unconditionally for years. ~40 LOC + 59 test LOC. *Verified: 537 OK.* **Keep the `transition_logger.error(...)` call** when inlining the enforce branch — `README:958` promises it is logged at ERROR and the harness reads that row.
> CHANGELOG: The phase-2 state guard now always enforces. A stale `PHASE2_STATE_GUARD: 'warn'` is ignored.

**3. `SENTRY_TRANSACTION_NAMING`** (`background/settings.py:257-262`) — always name the transaction. Unset in both trees. **Keep** `observability.py` and the duplicated `shadow` literal. ~10 LOC.

**All three are breaking, and silently so:** `DJANGO_LOGIC` has **no unknown-key rejection**, so a removed setting left in a consumer's config fails with no warning. Every removal needs a CHANGELOG "Removed" line. Also update the shared `test_unset_optional_settings_validate_clean` at `tests/background/test_settings_validation.py:231`.

Census after these three: 16 → 13 (→ 11-12 if `PROCESS_CLASS_ALIASES` also goes).

---
**KEEP `DEFER_UNLOCK_UNTIL_COMMIT`** (~130 engine + 319 test LOC) despite being the biggest never-set knob. It is pinned by a reproduction on **two real Postgres connections** (`tests/stability/test_transaction_integration.py::TestDeferUnlockTwoConnections` pins both the racy default and the fix) and was Heroku-validated as TC-27/TC-28, where the savepoint-release arm fixes a lock leak observable as a live Redis key. Consumers drive sync transitions inside `transaction.atomic()` at 11+ sites, so they are exposed today. "Nobody has opted in yet" is not speculative generality — the right follow-up is a *consumer* config decision.

### LINKED ISSUE #167: [1.0] PROCESS_CLASS_ALIASES is half-implemented — remove it or fix it
`PROCESS_CLASS_ALIASES` (`background/settings.py:186-206`, consumed at `runner.py:780-782,818`) is documented as the rename escape hatch paired with #140's fail-closed restore, and #149 boot-validates it. **It does not work as documented.**

`_find_background_transition_in_owner` compares `owning_process_class` **verbatim and never consults the alias map**. So with a correctly configured alias map:
- a simple Process rename still logs "renamed or removed?" and
- a nested condition-disambiguated one still yields `[unrestorable]`.

Its two existing tests pass only because they construct rows with a **blank** `owning_process_class` — a shape phase 1 has not produced since 0.4.x.

**A knob that does not do what its docstring claims is worse than no knob.** Two ways out; this needs a decision:

**(a) Remove it** (~35 LOC + 85 test LOC). Breaking, silent (no unknown-key rejection). Migration note must say plainly:
> A Process rename with in-flight rows now always fails closed — rows complete `[unrestorable]` and the stranded sweep drives the instance to `failed_state`. Drain in-flight rows before renaming a Process class.

That is already the documented rule for the sibling `action_name` refactor, so this collapses two rules into one.

**(b) Fix it** — apply the alias map in `_find_background_transition_in_owner` and rebuild the tests on realistic row shapes. This is an addition, so it was out of scope for the diet review.

One reviewer argued for keeping it (it is the paired drain path for #140 and #149 validates it); another *executed* the removal and measured both failure shapes above. Recommendation: **(a)**, with eyes open — but it is a product call.

Instance of a wider pattern worth naming: the #148/#149 hardening round added boot validators for two knobs **neither of which any consumer has ever set**. The cheap policy that would have prevented most of this: *don't validate, document or optimize a knob until a consumer sets it.* Note the #149 CHANGELOG entry names concrete reproduced bugs for the knobs consumers do set (negative `RETRY_MINUTES` hot-looped the starter; negative `CLEANUP_DAYS` deleted every completed row) and names no bug at all for the two nobody sets.

### LINKED ISSUE #168: [1.0] Remove conditions.py and get_transition_by_action_name — no callers anywhere
Two removals from the public surface, both with zero callers anywhere.

**1. `django_logic/conditions.py` (`all_related_in` / `any_related_in`)** — the whole 47-line module plus its test. The consumer implements this pattern **by hand four times** with *different* empty-set semantics, so the shipped helper is not what anyone actually wants; the harness usage is a retrofit that is reversible from git (~10-line revert available verbatim). ~86 LOC with tests.

Four coordinated edits: delete module + test, inline the closures into `docs/recipes/nested-processes.md:31,50-53`, CHANGELOG note, revert `fulfillment/processes.py` in the harness to its `actions.*` guards.

**2. `get_transition_by_action_name` + the public `ignore_state` parameter** (`process.py:206-207,256-258`) — zero callers in the library, its 13k-LOC suite, or either consumer. The docstring already describes `ignore_state` as "internal use". ~9 LOC. Use `_resolve_transition_with_owner` internally.

Note post-removal behaviour: an external caller gets a misleading `TypeError` rather than `AttributeError`, because `Process.__getattr__` treats any non-underscore miss as an action name. Repoint the 5 prose references or they dangle.

Both are breaking (public names on a published package) → CHANGELOG "Removed" lines, 1.0.

---
**Also in scope for a 1.0 sweep, lowest value — probably skip:** deleting `kwargs.pop('action_name')` at `process.py:123-127` (orphaned by 04d885a, which deleted the engine's own injection). It is 6 LOC for a public kwarg-contract narrowing on a surface where the consumer splats untrusted client JSON into this entrypoint across 152 view classes, so a currently-succeeding POST carrying `action_name` starts returning 400. Only do it if you want that loudness.

### LINKED ISSUE #169: Two stale guards: acks_late test list omits recover_stranded_states; dispatch docstring predates #151
Two stale guards found while verifying that other things were load-bearing. Both are small and both are real.

**1. `test_issue_fixes_engine.py:160-166` hard-codes the list of tasks that must declare `acks_late=True` + `reject_on_worker_lost=True`, and the list omits `recover_stranded_states`.** Consequence: you can strip **both** kwargs from that task and the full suite still passes. Those two kwargs are exactly what the crash-redelivery guarantee depends on (#91), so the test should derive the list from the task registry rather than hard-code it.

**2. `django_logic/background/dispatch.py` class docstring still claims `set_state` uses `xx=True`.** #151 moved the production path to a server-side Lua compare-and-set. One-line docstring fix.

Neither is overengineering — they are drift. Filing separately so they do not get lost inside the diet work.

### LINKED ISSUE #170: Add a system check for celery mode with the safety-net beat entries missing (cost a consumer 7 weeks of dropped retries)
### What happened

A consumer ran `BACKGROUND_EXECUTION='celery'` for ~7 weeks with **none of the five safety-net periodic tasks ever firing**, and nothing surfaced it.

Its `celery_app.py` did the documented thing:

```python
app.conf.beat_schedule = {**(app.conf.beat_schedule or {}), **beat_schedule()}
```

That is a silent no-op when the project also defines the Django-namespaced `CELERY_BEAT_SCHEDULE` setting. Celery's `app.conf` is a `Settings` ChainMap of `[0] runtime changes, [1] DictAttribute over Django settings, [2] defaults`; with `config_from_object('django.conf:settings', namespace='CELERY')` the **namespaced** key in MAP[1] wins over the `beat_schedule` key written into MAP[0]. Measured on the live app:

```
BASELINE                            24 entries, recover-stranded absent
app.conf.beat_schedule = {...}      24, absent   <- the documented pattern
app.conf['beat_schedule'] = {...}   24, absent
app.conf.update(beat_schedule=...)  24, absent
app.conf['CELERY_BEAT_SCHEDULE']    29, present  <- works
```

All five tasks were registered in the task registry the whole time — only never scheduled.

**Observed impact:** 6,612 `TransitionMessage` rows with **36 incomplete**, 5,184 rows past the cleanup window, oldest 46 days. 32 of the 36 had `errors_count=1` of `MAX_ERRORS=3` — retries the starter should have performed, aged up to **38.7 days**.

### Proposal

Add a `django_logic` system check (`W00x`, or `E00x`) that fires when `BACKGROUND_EXECUTION == 'celery'` but the entries returned by `background.beat_schedule()` are absent from the running app's `app.conf.beat_schedule`.

The engine already validates every safety *setting* at boot (#149). This is the same class of guarantee for the safety *net itself*, and it is currently the one precondition of celery mode that fails completely silently — the durability promise of `BACKGROUND_EXECUTION='celery'` is void without it.

Implementation notes:
- Needs the Celery app; resolve it lazily (`celery.current_app`) and skip the check cleanly when no app is configured, so sync-mode and library-only installs are unaffected.
- Worth pairing with a README note that assigning `app.conf.beat_schedule` is unreliable under `namespace='CELERY'`, and that consumers should write the namespaced key.

This is an addition rather than a diet item, filed because the failure mode is invisible and cost a consumer 7 weeks of dropped retries.

### LINKED ISSUE #171: [epic] Overengineering review of 0.9.1 — diet plan, and the KEEP list
Tracking issue for the whole-project overengineering review of 0.9.1.

### Verdict

**The library is overengineered, but not where its size suggests.** The *structure* is sound: 6,852 LOC across 30 modules, 540 tests in 4.1s, no committed build artifacts, and almost every guard that looks paranoid is pinned by a bug that was actually reproduced (#85, #87, #136, #138, #141, #143). 71 candidates were filed by six subsystem reviewers; adversarial verification killed or shrank **23** of them, and they mostly died because the "unnecessary" code turned out to be the remedy for a Heroku-reproduced incident.

The weight sits in three concentrations:

1. **A whole second State backend nobody uses** — `RedisState` is 199 of `state.py`'s 326 lines, and issues **#139** and **#151** (the top entry of the 0.9.1 release notes) hardened code with zero users. → #165
2. **16 settings, 6 never set by anyone** — each also carrying a boot validator, README lines and dedicated tests: roughly a 4x tax per knob. → #166, #167
3. **6,033 lines of Markdown for a 6,852-LOC engine**, ~2,300 of it pre-implementation plans that are now provably wrong. → #157, #162

**Available reduction:** ~1,050 engine LOC (≈15%), ~820 test LOC, ~2,600 doc lines, 16 knobs → 10-11, and ~10 concepts a new reader stops having to learn.

### DO NOW — 0.9.2, non-breaking (~2 half-days; six of these were executed against the suite by a verifier)

- [ ] #157 — delete 5 stale docs + `docs/INDEX.md` (~2,311 lines) — **do the pointer repointing listed in the issue first**
- [ ] #162 — prose/README de-duplication (**read the scope correction before starting**)
- [ ] #158 — delete `BaseTransition` + 4 unused `*_class` hooks
- [ ] #159 — delete the duplicated failure-side-effect savepoint layer
- [ ] #160 — delete `coverage.py` cross-version log readers
- [ ] #161 — delete the non-finite float pre-scan
- [ ] #163 — dead-code sweep, 8 small items
- [ ] #164 — test/CI/packaging cleanup

### 1.0 BREAKING DIET (~2 days + a release decision)

- [ ] #165 — remove `RedisState` (the big one; ship `RedisState = State` for one release as insurance)
- [ ] #166 — remove 3 never-set settings
- [ ] #167 — `PROCESS_CLASS_ALIASES`: remove or fix (needs a product call)
- [ ] #168 — remove `conditions.py` + `get_transition_by_action_name`

Every removed setting fails **silently** — `DJANGO_LOGIC` has no unknown-key rejection — so each needs a CHANGELOG "Removed" line.

### Also filed

- [ ] #169 — two stale guards (drift, not overengineering)
- [ ] #170 — new system check for celery mode with beat entries missing

### EXPLICITLY KEEP — do not re-litigate

Each of these looks excessive and is load-bearing. The bug each prevents:

| Thing | Why it stays |
|---|---|
| **`DEFER_UNLOCK_UNTIL_COMMIT`** (~130 + 319 test LOC), the biggest never-set knob | Reproduced on two real Postgres connections (`TestDeferUnlockTwoConnections`); Heroku-validated as TC-27/TC-28 where the savepoint arm fixes a lock leak observable as a live Redis key. Consumers drive sync transitions inside `atomic()` at 11+ sites, so they are exposed today. The follow-up is a *consumer config* decision, not a deletion. |
| **`timeout=` + `watchdog_stale_attempts` + `abandon_timed_out_attempt`** (~370 LOC) | The **only** mechanism that terminalizes an attempt dying without raising (OOM/SIGKILL/dyno cycle). `retry_stale_transitions` filters `errors_count < max_errors`, always true at 0, so it would redispatch forever; `detect_stuck` needs `>= max_errors` so it never fires; the stranded sweep excludes background rows. |
| **The five command bundles** | Five distinct contracts, not boilerplate. `Conditions`/`Permissions` have different signatures and consumers override the `user is None` fork in *both* directions; `SideEffects` is the only bundle driving `fail_transition`; `Callbacks`' per-item savepoint is #138; `FailureSideEffects` swallows-but-returns so phase 2 can record the error. |
| **`nested_processes` + `owning_process_class`** | #98. A consumer nests processes where both declare the same background `action_name`, separated only by conditions — without the recorded owner, phase 2 cannot pick the right one. 40 declarations across 31 files. |
| **Bind-time guards + `_validate_unique_*`** | #143: the model property was silently overwritten while the registry kept both claims, so every registry consumer disagreed with runtime dispatch. The guards are *coupled* — the idempotent re-bind early return is required by the conflict guard. |
| **Hook-signature validation on both surfaces** | One implementation, two messages. System checks do **not** run under WSGI or Celery workers, so bind-time is the only enforcement in a real web+worker process. Strict mode is what found three real offenders in a consumer tree showing zero warnings (#125). |
| **`E002` router-topology check** | The engine hardcodes the `default` alias; violating it commits the state write and the outbox row independently → stranded instances with no durable record. |
| **The stranded sweep + pk-keyset paging** | TC-24, reproduced on real dynos. The `instance_id__in` chunking is not a scale optimization — un-chunked it raises `too many SQL variables` and the sweep swallows it into a log line. |
| **`duration_ms` / `state_trace` / `assert_idempotent` / `ProcessScenario.snapshot()`** | Each has a named pinning test or live consumer call. `snapshot()` is the #150 fix. |
| **7-migration chain, `test_issue_fixes_*.py`, CHANGELOG, `docs/research/race-condition-issue`** | 0001-0005 all shipped in v0.3.0 — only two released schema states, nothing to squash. `test_issue_fixes_*` are mutation-verified: 12 of 13 reverts are caught by the test naming the issue. |

### The generalizable lesson

**Don't validate, document or optimize a knob until a consumer sets it.** #148/#149 added boot validators for `LOG_KWARGS_REDACTOR` and `PROCESS_CLASS_ALIASES` — neither has ever been set, and the latter is still half-implemented. `coverage.py` got upgrade-path readers for logs nobody persists. The timeout watchdog got a DB-floor optimization for a row volume nobody has. Note that #149's CHANGELOG entry names concrete reproduced bugs for the knobs consumers *do* set, and no bug at all for the two nobody sets.

### What is genuinely fine

`background/transitions.py`, `exceptions.py`, `apps.py`, `conf.py`, `commands.py`, `checks.py`, `background/models.py`, `observability.py`, the migration chain, and the test suite (13,449 test LOC : 6,852 engine LOC is a healthy 2:1). Repo hygiene is clean — `build/`, `egg-info/`, `__pycache__`, `.coverage` are all gitignored and uncommitted.

emil-balashov:
Works through every issue from the overengineering audit — #157–#164 (non-breaking) and #165–#170 (breaking + the one addition) — and cuts **0.10.0**. Closes the epic #171.

**−4,481 / +594 across 81 files.** Engine 6,852 → 6,305 LOC, docs −2,311 lines, settings 16 → 12 keys.

## Gate: verified safe against every open consumer PR first

Before deleting any public name I ran a compatibility scan over **all 46 open PRs in Borderless360/gv**, enumerating every consumer-side engine import under both spellings (`django_logic.*` and `django_logic_new.*`) — 22 distinct imports, all of which survive. Per-removal consumer count was **zero** for `RedisState`, `conditions.py`, `get_transition_by_action_name`, `PROCESS_CLASS_ALIASES`, `LOG_KWARGS*`, `PHASE2_STATE_GUARD='warn'`, `SENTRY_TRANSACTION_NAMING`, `snapshot_on_failure`, `ExecutionTracker`, the three testing helpers and `JourneyStep.matches()`.

Two mechanical consumer follow-ups, neither blocking: a gv test fixture builds a `TransitionMessage` with no `field_name` (add one), and a persisted coverage log needs re-measuring.

## Commits

| | |
|---|---|
| `4a10a97` | docs: five stale planning/research docs + the index that tiered them |
| `b108bf6` | chore: test-settings dedupe; three things that were never wired up |
| `de6a560` | refactor: engine code with no users and no reachable path |
| `7a61616` | **feat!**: remove `RedisState` |
| `d289690` | **feat!**: remove four never-set settings |
| `074a8e7` | **feat!**: remove public API with no callers |
| `17defd7` | fix: catch an unscheduled safety net; stop recommending the broken install |
| `5a17560` | release: 0.10.0 |

## The three findings worth your attention

**1. `RedisState` was 199 of `state.py`'s 326 lines, and #139 + #151 hardened dead code.** #151 — the Lua compare-and-set — is the top entry of the 0.9.1 release notes we just shipped. No consumer had ever set `state_class`; measured live on 28 bound machines / 137 transitions, all 50 process classes resolved to the base `State`. `state.py` is now 104 lines. The base-`State` half of #139 (ownership token, compare-and-delete unlock) is load-bearing and stays.

**2. `PROCESS_CLASS_ALIASES` never did what it documented.** It *is* applied to the recorded bound class, but `_find_background_transition_in_owner` compares `owning_process_class` verbatim and never consults the map — so renaming a *nested* owning process still yields `[unrestorable]` for a condition-disambiguated sibling set, with a correct alias configured. Its two tests passed only because they build rows with a blank `owning_process_class`, a shape phase 1 hasn't produced since 0.4.x. Removed rather than fixed; the migration note is "drain in-flight rows before renaming a Process class", which is already the rule for the sibling `action_name` refactor.

**3. We were documenting the install that silently fails.** Both the README and `beat_schedule()`'s own docstring said:

```python
app.conf.beat_schedule = {**app.conf.beat_schedule, **beat_schedule()}
```

Celery ignores that whenever the project also defines `CELERY_BEAT_SCHEDULE` in Django settings (`app.conf` is a ChainMap and the namespaced key wins). A consumer followed it and ran **seven weeks with all five safety-net tasks registered and never fired** — 36 stranded rows, 32 of them retries stuck at `errors_count=1` of 3 for up to 38.7 days, and a table that was never pruned. Both now recommend `app.conf['CELERY_BEAT_SCHEDULE']`, and **`django_logic.W002`** reports any shipped task missing from the running app's beat schedule. Matched by task name so renamed entries pass; a Warning, not an Error, because django-celery-beat's DB scheduler is legitimate — the hint names `SILENCED_SYSTEM_CHECKS`.

## Three places I deliberately deviated from the audit

- **Kept `BackgroundAction.complete_transition`.** The audit called it an unreachable no-op. But `BackgroundAction` extends `BackgroundTransition`, *not* `Action`, so it inherits `Transition.complete_transition`, which would write this action's empty target. Unreachable through the engine today ≠ unreachable on a public class.
- **Kept `_effective_timeout`** (inlined rather than deleted). The audit said it existed only to serve `RedisState`; base `lock()` calls it.
- **No `RedisState = State` alias.** The audit suggested one release of it as insurance. The two classes differ in *locking* semantics, so an alias changes behaviour silently where an `ImportError` says so plainly.

Also scoped down the prose de-duplication: the audit's ~400-line target rested on a thesis that verification measured as ~50× overstated (across `process.py` + `transition.py`'s ~490 prose lines, **8** carry an issue number). This does the defensible ~170 lines — blocks the README and CHANGELOG already own — and no density target. The `CLAUDE.md` comment rule that three reviewers cited as existing is now actually written down.

## Verification

- SQLite suite **488 OK**
- real-Redis suite (`tests.settings_redis`) **488 OK**
- **Postgres + Redis stability suite 52 OK** — the two-connection DEFER tests, deadlocks, race conditions, state consistency
- every `django_logic.*` module imports cleanly; documented public surface imports; `Transition.is_background` is still a public attribute (gv reads it via `getattr` with a default, so its loss would have been **silent** and would have reintroduced a `request`-in-background-kwargs regression)
- `manage.py check` clean; sdist + wheel build and ship none of the deleted modules

## Explicitly kept — please don't re-litigate

`DEFER_UNLOCK_UNTIL_COMMIT` (the largest never-set knob, but pinned by a two-real-Postgres-connection reproduction, and consumers drive sync transitions inside `atomic()` at 11+ sites — that one is a *consumer config* decision), the timeout watchdog, the five command bundles, `nested_processes` + `owning_process_class`, the bind-time guards, hook-signature validation on both surfaces, `E002`, the stranded sweep and its pk-keyset paging, `duration_ms`, `state_trace`, `assert_idempotent`, `ProcessScenario.snapshot()`, the 7-migration chain and `test_issue_fixes_*.py`. #171 records the bug each one prevents.

## Version

0.10.0, not 1.0.0 — pre-1.0, where a minor bump conventionally carries breaking changes. Say the word and I'll flip it; it's one line plus the CHANGELOG heading.

---

> [!NOTE]
> **High Risk**
> Wide breaking API and settings removals plus stricter TransitionMessage restore and unconditional phase-2 supersede behavior; upgrades need grep-driven migration and beat schedule verification via W002.
> 
> **Overview**
> **0.10.0** is a breaking cleanup release: large chunks of unused engine, docs, and CI scaffolding are removed, several public APIs and `DJANGO_LOGIC` keys are retired, and durability misconfiguration is easier to catch at `manage.py check`.
> 
> **Breaking removals** include **`RedisState`** and **`get_db_state()`** (use **`get_persisted_state()`**), **`django_logic.conditions`** (`all_related_in` / `any_related_in`), **`Process.get_transition_by_action_name`** and **`get_available_transitions(..., ignore_state=)`**, **`PROCESS_CLASS_ALIASES`**, **`LOG_KWARGS` / `LOG_KWARGS_REDACTOR`**, **`PHASE2_STATE_GUARD='warn'`** (guard always enforces supersede), **`SENTRY_TRANSACTION_NAMING`**, **`ProcessScenario.snapshot_on_failure`**, **`JourneyStep.matches()`** (use `==`), **`ExecutionTracker`** re-export, and folding **`BaseTransition`** into **`Transition`**. **`TransitionMessage`** restore now **requires `field_name`** on testing helpers and phase-2 restore (no `'state'` guess). Transition coverage logs no longer accept legacy 0.8/0.9.0 key formats.
> 
> **Behavior fixes / simplification:** phase-2 and safety-net **`failed_state`** writes always respect the state guard; failure side-effects go through **`FailureSideEffects.execute`** only (duplicate savepoint wrapper removed). Sentry naming is always on when `sentry-sdk` is present.
> 
> **Added:** **`django_logic.W002`** warns when celery mode is on but the five safety-net beat tasks are missing from the schedule; **`django_logic.W003`** warns on removed `DJANGO_LOGIC` keys still present in settings. Docs now tell consumers to set **`app.conf['CELERY_BEAT_SCHEDULE']`**, not `app.conf.beat_schedule`, when `CELERY_BEAT_SCHEDULE` exists in Django settings.
> 
> **Repo/docs:** stale planning docs and `.cursor/rules` removed; README/CLAUDE/testing guide trimmed; nightly stability “stress” CI step, broken compose service, and unused dev deps dropped; test settings consolidated via **`tests.dl_settings`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88d28f9f201b6cde8c6d0d16e64323dcece397f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

## Review round (2026-07-27)

All six pre-merge items from the release-readiness review are addressed in `88d28f9`.

1. **W002 was missing E002's installed-app gate** — the one real defect. `BACKGROUND_EXECUTION` defaults to `'celery'` and the core app registers checks too, so an install without `django_logic.background` got a false warning on every `manage.py check`, and would fail CI running `check --fail-level WARNING`. Fixed with the same `apps.is_installed` gate, plus a test that removes the app from `INSTALLED_APPS`.
2. **CHANGELOG** — the `get_transition_by_action_name` entry now names the replacement (`get_available_transitions(action_name=...)`) and quotes the real error, verified empirically rather than asserted: `Process.__getattr__` treats the stale name as an *action*, so you get `TypeError: get_transition_by_action_name() accepts keyword arguments only (got 1 positional)`, not an `AttributeError` pointing at the removal.
3. **Stale docs** — `CLAUDE.md`'s `PHASE2_STATE_GUARD` mention, `docs/logger.md`'s `__redaction_error__` fallback, and the `pyproject.toml` comment justifying `django-redis` via `RedisState` are all fixed. The dependency question itself is filed as **#173** rather than decided inside a diet PR: the engine has never imported `django_redis`, and Django has shipped its own Redis backend since 4.0, but demoting it changes the install surface and deserves its own release note.

   I also found a **fourth site the review didn't flag**: `CLAUDE.md:81` recommended `app.conf.beat_schedule = beat_schedule()` — the exact broken install W002 exists to catch. Fixed.
4. **Stale tests** — deleted `test_observability.test_disabled_via_setting` (overrode a removed setting, so it was an unfailable duplicate of `test_no_op_without_sentry_sdk`); rewrote `test_log_redaction` around the shallow copy that survives and dropped the dead `drop_amount`/`_boom` helpers; removed `'warn'` from `test_phase2_state_guard`'s docstring.
5. **Issue closing** — fixed below. All fifteen are fully implemented, so all fifteen close.
6. **The fail-open judgment call — decided: warn, don't stay silent.** Added **`django_logic.W003`**, which reports any `DJANGO_LOGIC` key this release removed and names its replacement. The reviewer's own sharpest case is why: a deployment that set `LOG_KWARGS_REDACTOR` for PII compliance upgrades and starts writing raw kwargs to its logs with no signal anywhere. "Zero known consumers set it" was the right criterion for *removing* the knob; it is not a good reason to make the upgrade silent for anyone outside that sample. ~40 lines and 5 tests.

**Re-verified after these changes:** sqlite **493 OK**, real-Redis **493 OK**, Postgres+Redis stability **52 OK**, `manage.py check` clean.

### Consumer evidence (RELEASING.md step 6)

The `django-logic-test` Heroku rig passed against this candidate — recorded in that repo's `docs/RESULTS.md` at `b0377b3`:

- **75 on-dyno pytest, 0 skipped**; **7 TC-26/27/28 rows against real Heroku Redis** (`rediss://`); **25/25 across every matrix group** (smoke 3, p0 9, concurrency 3, durability 3, p1 5, tc06 2).
- That run caught two things worth repeating here. The TC-26/27/28 guards probed `hasattr(State, '_stored_token')` as a proxy for "engine has #139/#141" — this PR deletes `_stored_token` with `RedisState`, so six tests silently stopped running while the suite still reported green and the dyno still exited 0. And reverting the harness to its pre-#81 hand-written parent guards fixed a latent bug: `all_related_in` used `not exclude(...).exists()`, **True for an empty set**, so a childless order satisfied `mark_fulfilled`. Differing empty-set semantics across consumers is exactly why this PR drops the helper.

The gv-side re-vendor and its coverage-log re-measure remain the one follow-up outside either repo.

---

Closes #157, closes #158, closes #159, closes #160, closes #161, closes #162, closes #163, closes #164, closes #165, closes #166, closes #167, closes #168, closes #169, closes #170, closes #171


emil-balashov:
## Release-readiness review — 0.10.0

Ran a full review of this PR across six dimensions (RELEASING.md checklist, removal completeness, new/refactored code, behavior edges, consumer compat against the real gv tree, packaging dry-run), with every finding independently re-verified against the code. **No blockers — this is ready to merge once the small items below are decided.** What's left before *publish* is the usual step‑6 consumer evidence.

### Verified ✅

- CI 14/14 green; suite re-run locally in the PR worktree: **488 OK** including the `tests.test_metadata` drift check (py3.12 / Django 6.0.4).
- `pyproject.toml` at **0.10.0**; CHANGELOG has the dated `[0.10.0] — 2026-07-27` section with prose overview + a fresh empty `[Unreleased]`; README is clean of every removed API/setting and the support statement matches the CI matrix.
- Every breaking removal is actually gone — no dangling imports, `__init__`/`__all__` all resolve, the message helpers require `process_name` at every call site.
- **W002**: name-matched (renamed entries pass), covers all five safety-net tasks, Warning with the `SILENCED_SYSTEM_CHECKS` hint, exception-safe when Celery config is lazy/absent; the `app.conf.beat_schedule`-shadowing claim was reproduced empirically against celery 5.6.3.
- `FailureSideEffects` consolidation: the removed runner wrapper was a strictly redundant second savepoint; the phase-2 enforce path is byte-identical to master (only `warn` branches deleted, default was already `enforce`).
- `field_name` fail-closed: tested, contained ( `[unrestorable]` → stranded recovery), and **not** a 0.9.x data hazard — phase 1 has recorded `field_name` since 0.4. Base `State` keeps #139's ownership token + compare-and-delete unlock; `_effective_timeout` inlining is semantics-identical; the `BackgroundAction.complete_transition` keep-rationale holds (it shadows `Transition.complete_transition`, which would write the action's empty target).
- **Consumer scan against the real gv checkout**: every name gv touches survives — 9 module surfaces, all 21 `ProcessScenario`/assertion helpers across its 15 journey-test files, `bind_model_process` signature, all seven `DJANGO_LOGIC_NEW` settings keys, identical 7-migration chain — and zero usage of anything removed. Both disclosed follow-ups check out: the `field_name`-less fixtures (`rate/tests/test_rate_bundle_background.py:134,158`) would in fact still pass under 0.10.0. gv master already carries the namespaced beat-key fix (`fbba85c16`), so W002 will pass there on re-vendor.
- **Packaging dry-run** (steps 7 + 11 equivalent): `uv build` → 0.10.0 sdist + wheel, `twine check` PASS, none of the deleted modules ship / all 38 surviving ones do, clean-venv install + full documented-surface import OK, `[celery,redis]` extras still resolve.

### Before merge — small fixes to decide ☐

1. **`django_logic/checks.py:158` — W002 is missing E002's installed-app gate.** E002 returns early via `apps.is_installed('django_logic.background')` (checks.py:107); W002 doesn't. The core app also registers checks (`django_logic/apps.py:17`), `BACKGROUND_EXECUTION` defaults to `'celery'`, and celery is a hard dependency — so an install *without* the background app gets a false W002 on every `manage.py check`, and fails any CI running `check --fail-level WARNING`. One-line gate.
2. **`CHANGELOG.md:43`** — the `get_transition_by_action_name` / `ignore_state=` entry is the only removal in the section with no replacement or rationale. Worth naming `get_available_transitions(...)` and surfacing the gotcha from 074a8e7's own commit message: a stale call now raises a misleading `TypeError` via `Process.__getattr__`, not an `AttributeError` pointing at the removal.
3. **Stale docs the sweep left behind**: `CLAUDE.md:68` still documents `DJANGO_LOGIC['PHASE2_STATE_GUARD']` as current; `docs/logger.md:38` still describes the removed `__redaction_error__` redactor fallback; `pyproject.toml:45`'s comment still justifies the `django-redis` dependency via RedisState — and the package no longer imports `django_redis` at all, so decide whether the core dep itself stays.
4. **Three vacuous/stale tests**: `tests/background/test_observability.py:32` (`test_disabled_via_setting` overrides the removed `SENTRY_TRANSACTION_NAMING` — now an unfailable duplicate of `test_no_op_without_sentry_sdk`); `tests/test_log_redaction.py:23` (wired to the removed `LOG_KWARGS_REDACTOR`, plus the dead `_boom` helper); `tests/background/test_phase2_state_guard.py:8` (docstring still specifies the deleted `'warn'` mode).
5. **Issue closing**: GitHub registers **zero** closing references on this PR — "Closes the epic #171" doesn't parse as a closing keyword. Commit messages auto-close #165–#170 only; #157–#164 are `Refs`. Either add a bare `Closes #171` (and decide whether #157–#164 should close too) to the body before merge, or close them manually after.
6. **Judgment call, fine to ship as-is**: all four removed settings fail *open* silently at boot. Sharpest edge: a deployment that had set `LOG_KWARGS`/`LOG_KWARGS_REDACTOR` for PII redaction upgrades and starts logging raw kwargs with no signal. Zero known consumers set them (that was the removal criterion), so the CHANGELOG note is probably enough — flagging so it's a decision, not an accident.

### Before publish — RELEASING.md steps 6–11 ☐

- **Step 6 consumer evidence** (0.9.x precedent: posted on the release PR before merge):
  - gv suite with the candidate re-vendored as `django_logic_new` — the legacy-format coverage-log re-measure is the one follow-up living outside either repo.
  - `django-logic-test` Heroku rig pass against the candidate (usual throwaway-commit pin procedure).
- Steps 7–11 from `master` after merge: `make dist` → `make publish` → tag `v0.10.0` → GH release with the changelog section + artifacts → clean-venv `pip install django-logic==0.10.0`. Step 7 was dry-run-verified above; rerun from master.

emil-balashov:
All six pre-merge items are addressed in `88d28f9`. Thanks — item 1 was a genuine defect and item 6 changed my mind.

**1. W002's missing installed-app gate — fixed.** You're right that this was the real one: `BACKGROUND_EXECUTION` defaults to `'celery'`, the core app registers checks too, and celery is a hard dependency, so an install without `django_logic.background` got a false warning on every `manage.py check` and would fail `check --fail-level WARNING`. Same `apps.is_installed` gate as E002, plus a test that removes the app from `INSTALLED_APPS` so the gate can't silently regress.

**2. CHANGELOG — fixed, and I verified the gotcha rather than paraphrasing the commit message.** It names `get_available_transitions(action_name=...)` as the replacement and quotes what actually happens: `Process.__getattr__` treats the stale name as an *action*, so attribute access still returns a callable and calling it reports

```
TypeError: get_transition_by_action_name() accepts keyword arguments only (got 1 positional). Pass user and other values
```

which reads like an argument mistake on an existing action, not a removal. The entry says to grep for the name rather than trust the traceback.

**3. Stale docs — all three fixed, plus a fourth you didn't flag.** `CLAUDE.md`'s `PHASE2_STATE_GUARD` mention, `docs/logger.md`'s `__redaction_error__` fallback, and the `pyproject.toml` comment are done. The fourth: **`CLAUDE.md:81` recommended `app.conf.beat_schedule = beat_schedule()`** — the exact broken install W002 exists to catch. Two of the three places we documented the failing pattern were fixed in this PR and that one survived, which is a decent argument for W002 existing at all.

On the dependency itself: filed as **#173** rather than decided here. The engine has never imported `django_redis` (it locks through Django's cache API), and Django has shipped its own Redis backend since 4.0 while our floor is 4.2 — so the *stated* reason is gone. But it still needs a **cross-process** cache, demoting it breaks anyone relying on transitive installation, and the failure mode would be an `InvalidCacheBackendError` the engine can't explain. That's an install-surface change deserving its own release note, not a rider on a diet PR.

**4. Stale tests — all three fixed.** `test_disabled_via_setting` deleted (it was an unfailable duplicate of `test_no_op_without_sentry_sdk` once the setting was gone); `test_log_redaction` rewritten around the shallow copy that survives — it now actually pins the reason the copy exists, that later kwargs mutation must not leak into an already-emitted record — with the dead `drop_amount`/`_boom` helpers dropped; `'warn'` out of `test_phase2_state_guard`'s docstring.

**5. Issue closing — fixed.** Worth recording *why* it looked fine and wasn't: a comma-separated `closes #157, closes #158, …` list only registered **two** of the fifteen. One keyword per line registers all fifteen, which `gh pr view --json closingIssuesReferences` now confirms. All fifteen are fully implemented, so all fifteen close.

**6. The fail-open call — I've decided to warn rather than ship silent.** Added **`django_logic.W003`**, which reports any `DJANGO_LOGIC` key this release removed and names its replacement. Your framing is what changed my mind: "zero known consumers set them" was the right criterion for *removing* the knobs, but it is not a good reason to make the upgrade silent for anyone outside the sample we could see. The `LOG_KWARGS_REDACTOR` case is a deployment that opted into redaction for compliance, upgrades, and starts writing raw kwargs to its logs with nothing anywhere saying so — a CHANGELOG line only helps someone who reads it before deploying. ~40 lines, 5 tests, and it's the same lesson as W002: this release's whole theme is that silent failure is the expensive kind.

---

**Re-verified after all of the above:** sqlite **493 OK**, real-Redis **493 OK**, Postgres+Redis stability **52 OK**, `manage.py check` clean, CI green on `88d28f9` (13/13 reporting so far, `stability (3.12)` still running).

**Step-6 consumer evidence is in the PR body now** — the `django-logic-test` rig passed against this candidate (75 on-dyno pytest with 0 skipped, 7 TC-26/27/28 rows against real Heroku Redis, 25/25 across every matrix group; recorded at `b0377b3`). Two findings from that run are worth your attention because they're the same class of problem as your item 6:

- The harness's TC-26/27/28 guards probed `hasattr(State, '_stored_token')` as a proxy for "engine has #139/#141". This PR deletes `_stored_token` with `RedisState`, so **six tests silently stopped running** while the suite still reported green and the dyno still exited 0. Probing an implementation detail is how a green harness lies.
- Reverting the harness to its pre-#81 hand-written parent guards **fixed a latent bug**: `all_related_in` used `not exclude(...).exists()`, which is `True` for an empty set, so an order with *no children* satisfied `mark_fulfilled`. Differing empty-set semantics across consumers is exactly why this PR drops the helper.

Remaining before publish is what you listed: the gv re-vendor + coverage-log re-measure, then steps 7–11 from `master`. Ready to merge from my side — and still open for you: **0.10.0 vs 1.0.0**, which is one line plus the CHANGELOG heading.

emil-balashov:
## 🚀 0.10.0 released

- **PyPI:** https://pypi.org/project/django-logic/0.10.0/
- **GitHub release:** https://github.com/Borderless360/django-logic/releases/tag/v0.10.0
- **Tag:** `v0.10.0` → `0448766`

The `88d28f9` follow-up landed everything from the readiness review (W002 gate, CHANGELOG guidance, stale docs/tests) and added **W003**, which reports removed `DJANGO_LOGIC` keys instead of ignoring them — that closes the one item the review left as a judgment call.

### RELEASING.md, step by step

| Step | Result |
|---|---|
| 4 — tests | **493 OK** (SQLite) · **493 OK** (real Redis, `tests.settings_redis`) · Postgres+Redis stability green in CI on `0448766` |
| 5 — metadata drift | **4 OK** |
| 6 — consumer checks | see below |
| 7 — `make dist` | sdist + wheel built, `twine check` **PASSED** on both |
| 8 — `make publish` | uploaded |
| 9 — tag | `v0.10.0` pushed |
| 10 — GH release | created, both artifacts attached |
| 11 — clean-venv install | `pip install django-logic[celery,redis]==0.10.0` → imports clean |

### Step 6 — consumer validation (gv)

gv re-vendored the candidate at `a9ac1ba6d`. Verified against the real checkout (branch `staging`):

- **Vendor drift gate: 0 stale, 0 missing, 0 extra** across all 37 files — the vendored tree is an exact transformed mirror of `origin/master`, `exceptions.py` bridge excepted.
- `manage.py check` → **no issues**; no W002, W003 or E002.
- **`rate`** (the only app on the new engine): **470 OK**
- 14 other engine-touching apps: **1,063 OK**
- All engine-touching modules in `api`/`gv` (incl. `gv.tests.test_celery_beat_schedule`): **244 OK**
- **W002 under real celery-mode wiring:** simulated gv's `install_engine_beat_schedule` with `BACKGROUND_EXECUTION='celery'` — all 5 shipped safety-net tasks resolve in beat, **0 missing, no check fires**. The seven-week incident this check exists to catch is genuinely wired shut in gv.

One thing worth recording, since it briefly looked like a regression: running gv's full `api`+`gv` suites under `--parallel=4` produces ~120 failures. I A/B'd it on an isolated worktree — same `staging` commit, **only** the vendored package reverted to 0.9.1 — and got 117 failures there vs 115 on 0.10.0, with the two failure sets **almost entirely disjoint** (121 unique vs 114 unique, 3 shared). Pre-existing parallel-run flakiness in that local environment (shared Redis, missing Google creds), not a 0.10.0 effect — and every one of those tests passes serially. No failure in either run mentions `django_logic_new`.

### Clean-install behaviour verified from PyPI

Documented public surface imports; every removed name confirmed absent (`RedisState`, `State.get_db_state`, `conditions`, `get_transition_by_action_name`, `ExecutionTracker`, `BaseTransition`); both new checks registered; **no check messages on a clean install**; W003 fires correctly when a removed key is set; `beat_schedule()` ships 5 tasks.

### Follow-ups

- #173 — whether `django-redis` stays a core dependency now that the engine never imports it.
- gv-side: the persisted legacy-format coverage log still needs re-measuring (ops artifact, not in the checkout).
- The `[0.10.0]` CHANGELOG heading is dated 2026-07-27 (when the section was finalised); publish happened 2026-07-28.

PR #174, opened 2026-07-28
deps: drop django-redis from core, give the [redis] extra meaning again

### LINKED ISSUE #173: Decide whether django-redis stays a core dependency (the engine never imports it)
Surfaced while reviewing #172: the `pyproject.toml` comment justified `django-redis` as a core dependency via "`RedisState` / the lock cache", and 0.10.0 removed `RedisState`. The comment is corrected there, but the underlying question is worth deciding on its own.

**The engine has never imported `django_redis`.** It locks through Django's cache API (`cache.add` / `cache.get` / `cache.delete`), so nothing in `django_logic/` references the package:

```
$ grep -rn "django_redis" django_logic/ --include="*.py"
(no matches)
```

What the engine *does* require is a **cross-process** `default` cache — the state lock is worthless if web and workers each hold a private locmem dict, which is why celery mode refuses to boot on locmem/dummy when `DEBUG=False`. `django-redis` ships installed so that requirement is satisfied by default rather than becoming a footgun.

**What changed underneath us:** Django has had its own Redis backend since 4.0 (`django.core.cache.backends.redis.RedisCache`), and our floor is `django>=4.2`. So the *documented* way to get a cross-process cache no longer needs a third-party package.

### The call

- **Keep it** (status quo, what 0.10.0 ships): one transitive dependency, and every existing consumer whose settings name `django_redis.cache.RedisCache` keeps working untouched.
- **Demote to an extra** (`django-logic[redis]` — note `[redis]` currently exists as an empty alias, so this would give it meaning again): honest about what the engine actually imports, but it breaks any consumer relying on transitive installation, and the failure mode is an `InvalidCacheBackendError` at boot rather than something the engine can explain. Would want a release note and probably a check that says "your `default` cache backend isn't importable" more helpfully.

Not urgent and deliberately out of scope for the 0.10.0 sweep, which was about removing code with no users rather than changing the install surface. Worth doing in the same release as any other dependency change.

emil-balashov:
Closes #173.

Audited every declared dependency for whether the engine actually needs it. One core dependency doesn't survive the audit; the rest do, and I've recorded why so this isn't re-litigated.

## `django-redis` — removed from core, `[redis]` extra given meaning again

**The engine has never imported `django_redis`.** The state lock goes through Django's cache API — `cache.add()` to acquire, `cache.get()`/`cache.delete()` for the compare-and-delete release ([state.py:58-100](../blob/master/django_logic/state.py#L58-L100)). So what it requires is a **cross-process cache backend**, not a particular package. Django has shipped `django.core.cache.backends.redis.RedisCache` since 4.0, comfortably under our `>=4.2` floor.

### Locking is proven intact, not argued

Clean venv, `django-redis` uninstalled, Django's built-in backend, two separate OS processes on one key:

```
### Django built-in RedisCache (django-redis NOT installed)
proc1 acquire : ACQUIRED
proc2 acquire : BLOCKED
proc3 islocked: LOCKED
proc4 unlock  : RELEASED
proc5 acquire : ACQUIRED

### LocMemCache control (expected: no cross-process locking)
proc1 acquire : ACQUIRED
proc2 acquire : ACQUIRED
```

The locmem control is the point: it shows both processes acquiring, which is exactly the failure `_check_lock_cache_in_celery_mode()` exists to prevent. **That guard is unchanged and always was backend-agnostic** — it string-matches the `BACKEND` path against locmem/dummy. It never mentioned a package except in its error text, which this PR corrects.

### Consumer impact: none

Both known consumers already declare `django-redis` themselves rather than relying on the transitive install — gv (`pyproject.toml:47`, `django-redis >= 5.4.0, < 6`) and the `django-logic-test` rig (`requirements.txt:23`). Anyone naming `django_redis.cache.RedisCache` in settings keeps working.

`[redis]` stops being an empty alias and installs `django-redis`, so that pin does what its name says again. Verified against the built wheel:

| install | `django-redis` |
|---|---|
| `django_logic-0.10.0.whl` | absent ✅ |
| `…whl[redis]` | 7.0.0 + `redis` 8.0.1 ✅ |
| `…whl[celery,redis]` (legacy pin) | resolves, installs it ✅ |

**One sharp edge worth reading before merge:** a consumer who *did* rely on the transitive dep does **not** fail at boot. `django.setup()` succeeds and `InvalidCacheBackendError` fires at the first cache access — inside a request or a worker — because the guard never instantiates the cache. That's in the CHANGELOG migration note.

## Deliberately kept — with reasons

- **`celery`** — technically demotable, but I recommend against it and did not do it. `background/tasks.py:28` imports `shared_task` at module level and `dispatch.py:134` (`retry_pending`) imports from it unconditionally. Worse, `BACKGROUND_EXECUTION` defaults to `'celery'` and W002's `ImportError` guard means `manage.py check` would report **no issues** with celery absent — so a consumer who forgot the extra boots clean and silently strands their first background transition. Trading a hard `ImportError` for that is a bad deal.
- **`django-model-utils`** — genuinely used (`TimeStampedModel`), and `background/migrations/0001_initial.py:2` imports `model_utils.fields` at module level, so uninstalling breaks `migrate` before any model is touched. Removing it means either vendoring the field classes or rewriting an applied migration. Not worth it for one small, stable dependency.
- **`[dev]`** — `psycopg[binary]` is required by `stability.yml`, `coverage` by `make coverage`. Added `redis>=4.2` here, since the test settings now use Django's built-in backend.

## One open question for you

The **`drf` extra** installs `djangorestframework` and **nothing in the repo references `rest_framework`** — zero hits across code, tests and docs. I left it alone rather than deleting it, because [README.md:42](../blob/master/README.md#L42) documents it as an intentional shim ("kept for projects migrating off the old DRF-coupled releases") and `TODO.md:9` lists "Admin + DRF integration modules" as planned work. It's opt-in, so it costs nothing by default. Say the word and it's a two-line deletion.

## Changes

- `pyproject.toml` — `django-redis` out of core; `redis = ["django-redis>=5.0.0"]`; `redis>=4.2` into `[dev]`; comments rewritten
- `tests/settings_redis.py`, `tests/settings_stability.py` — Django's built-in Redis backend, so the suites no longer need the third-party package
- `django_logic/background/settings.py:270` — the guard message no longer claims django-redis ships as a dependency
- `README.md`, `CLAUDE.md` — install/deployment docs
- `CHANGELOG.md` — `[Unreleased]`, breaking, with the migration note

Version left at 0.10.0 — this is breaking for anyone relying on the transitive install, so it wants a 0.11.0 release PR rather than a version bump smuggled in here.

## Verification

- SQLite suite **493 OK**
- real-Redis suite on the built-in backend **493 OK**
- Postgres + Redis stability **52 OK**
- metadata drift **4 OK**
- wheel builds; all three install modes verified above

---

> [!NOTE]
> **Medium Risk**
> Breaking for installs that relied on transitive django-redis without declaring it; misconfigured cache backends may fail at first cache use rather than at install time.
> 
> **Overview**
> **Breaking dependency change (0.11.0):** `django-redis` is removed from core `dependencies` and moved to the optional `[redis]` extra (`pip install django-logic[redis]`). Celery stays required; only the Redis cache *client package* is no longer pulled in by default.
> 
> The library still locks only through Django’s cache API, so deployments can use **`django.core.cache.backends.redis.RedisCache`** with a `redis` client instead of `django_redis.cache.RedisCache`. Docs (`README`, `CLAUDE.md`, `CHANGELOG`) and the celery-mode locmem guard message in `django_logic/background/settings.py` are updated to say that. **Test settings** (`settings_redis.py`, `settings_stability.py`) now use the built-in Redis backend; **`redis>=4.2`** is added under the `dev` extra for those suites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0de1131812d7ccf30f49cc6cb03ca11433a19255. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 ## [Unreleased]
 
+### Changed (breaking)
+
+- **`django-redis` is no longer a core dependency** — it moves to the
+  `[redis]` extra, which stops being an empty alias (#173). The engine has
+  never imported `django_redis`: the state lock goes through Django's cache
+  API (`State.lock` → `cache.add`), so what it actually requires is a
+  *cross-process cache backend*, not a particular package. Django has shipped
+  `django.core.cache.backends.redis.RedisCache` since 4.0 and our floor is
+  4.2, so the documented setup no longer needs a third-party package.
+
+  Verified: with `django-redis` uninstalled and the built-in backend
+  configured, two separate OS processes contend correctly on the same lock —
+  the second `lock()` returns `False`, `is_locked()` is `True`, and the key
+  is released on `unlock()`. The locmem control shows both processes
+  acquiring, which is the failure the boot guard exists to prevent.
+
+  **Migration:** if your settings name `django_redis.cache.RedisCache`,
+  install it explicitly — `pip install django-logic[redis]`, or add
+  `django-redis` to your own dependencies. Both known consumers already
+  declare it directly and are unaffected. Note the failure mode if you miss
+  this is *not* at boot: `django.setup()` succeeds and
+  `InvalidCacheBackendError` is raised at the first cache access, because
+  the celery-mode guard only string-matches the backend path against
+  locmem/dummy and never instantiates the cache.
+
+  Not affected: the boot guard itself, which is backend-agnostic and still
+  refuses locmem/dummy in celery mode when `DEBUG=False`.
+
 ## [0.10.0] — 2026-07-27
 
 An overengineering sweep. ~1,900 lines removed across the engine, its tests and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,11 +69,14 @@ change on success) for anything slow, external, or retriable.
 
 ## Deployment the durability contract depends on
 
-- A real broker (Redis/RabbitMQ). Celery and django-redis are core
-  dependencies of django-logic (installed automatically);
-  `BACKGROUND_EXECUTION` defaults to `'celery'`.
-- A cross-process `default` cache (django-redis) for the state lock —
-  celery mode refuses to boot with a locmem/dummy cache when `DEBUG=False`.
+- A real broker (Redis/RabbitMQ). Celery is a core dependency of
+  django-logic (installed automatically); `BACKGROUND_EXECUTION` defaults
+  to `'celery'`.
+- A cross-process `default` cache for the state lock — celery mode refuses
+  to boot with a locmem/dummy cache when `DEBUG=False`. The engine locks
+  through Django's cache API and imports no backend, so Django's built-in
+  `django.core.cache.backends.redis.RedisCache` is enough; django-redis is
+  the `[redis]` extra, not a core dependency (0.11.0).
 - Crash re-delivery is built in (every django-logic task sets
   `acks_late=True` + `reject_on_worker_lost=True`); set the global Celery
   pair only for your *own* tasks. You still need a **single beat**
--- a/README.md
+++ b/README.md
@@ -35,17 +35,18 @@ Django Logic is a lightweight workflow framework for Django that makes it easy t
 - Django 4.2+ (4.2, 5.1, 5.2 and 6.0 are tested in CI; 5.0 is not supported)
 - django-model-utils >= 4.5.1
 - celery >= 5.0 — **installed automatically**; background transitions are Celery tasks
-- django-redis >= 5.0.0 — **installed automatically**; provides the cross-process state lock
+- A **cross-process `default` cache** for the state lock — *not* a package dependency. The engine locks through Django's cache API, so any cross-process backend works, including `django.core.cache.backends.redis.RedisCache`, built into Django since 4.0. Celery mode refuses to boot on a locmem/dummy cache when `DEBUG=False`.
 
 Extras:
+- `pip install django-logic[redis]` — installs `django-redis`, for deployments whose settings name `django_redis.cache.RedisCache`. It stopped being a core dependency in 0.11.0: the engine has never imported it.
 - `pip install django-logic[drf]` — pulls in `djangorestframework` (kept for projects migrating off the old DRF-coupled releases; no DRF-specific code ships since 0.4)
-- `[celery]` and `[redis]` remain as **empty aliases**, so existing `pip install django-logic[celery,redis]` pins keep resolving — both packages are core dependencies since 0.4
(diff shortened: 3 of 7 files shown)


emil-balashov:
## Self-review after #176 appeared — two coordination items

Re-checked this branch now that #176 is open. The dependency work itself still stands: CI 14/14 including both stability jobs, which are the meaningful signal here since they install `.[dev]` in a clean environment and exercise the changed cache settings. `Dockerfile` installs `.[dev]` too, so `make test` (locmem) and `make stability-redis` (built-in Redis backend) both keep working. No further code concerns from my side.

Two things that only became visible with the other PR in flight:

### 1. CHANGELOG conflicts with #176 — sequencing needs a decision

Both PRs insert directly under `## [Unreleased]`. GitHub marks both MERGEABLE because each is clean against master *independently*; merging either one makes the other conflict. I tested it — `CONFLICT (content): Merge conflict in CHANGELOG.md`, while `README.md`, `CLAUDE.md` and `pyproject.toml` auto-merge fine.

Trivial to resolve, but the release shape isn't:

- **#176 first** → publish 0.10.1, then rebase this and cut 0.11.0. Two releases, one story each. #176 already carries `version = "0.10.1"` and a dated heading, so this is the path of least edits.
- **Both, one 0.11.0** → #176's version bump and `[0.10.1]` heading fold into 0.11.0.

My preference is #176 first: it's a self-contained relaxation that breaks nothing, so it earns a patch release, whereas this one is breaking and wants the minor bump.

### 2. This branch forward-references 0.11.0 in three places

I wrote the version number into prose before the release number was settled:

- `README.md:41` — "It stopped being a core dependency in 0.11.0"
- `CLAUDE.md:78` — "not a core dependency (0.11.0)"
- `pyproject.toml:53` — "`[redis]` has meaning again as of 0.11.0"

(plus the `[Unreleased]` CHANGELOG section, which is correct by construction). `version` is deliberately left at 0.10.0 here so the release PR owns the bump — but if this ends up shipping as anything other than 0.11.0, those three strings are wrong and nothing will catch them, since `tests.test_metadata` checks Django/Python floors, not prose. Whoever cuts the release should grep `0.11.0` and confirm.

Given the breaking nature — a consumer relying on the transitive install gets `InvalidCacheBackendError` at first cache access, not at boot — 0.11.0 remains the right target under the repo's pre-1.0 convention.

Issue #143, opened 2026-07-22
Validate process bindings to prevent ambiguous stranded-state recovery


emil-balashov:
## Problem

`recover_stranded_states()` in PR #137 infers ownership from a model, state field, and `in_progress_state`, but uniqueness is enforced only inside one process tree. `ProcessManager.bind_model_process()` permits multiple process classes to bind to the same model field and does not validate conflicts globally.

If two bound machines share an `in_progress_state` on one field, a record-less synchronous stranding has no durable provenance. The first sweep can run the wrong transition's `failed_state` and failure hooks. The current `seen` key only changes which transition wins; it cannot prove ownership.

The same binding API also:
- silently overwrites an existing model property when `process_name` collides;
- keeps contradictory entries in `ProcessManager.bindings`;
- does not validate that `state_field` is a concrete model field.

Affected code:
- `django_logic/process.py` (`ProcessManager.bind_model_process`)
- `django_logic/background/dispatch.py` (`recover_stranded_states`, `_sweep_transition`)
- PR #137

## Impact

An invalid but currently accepted binding topology can execute the wrong compensation hooks and move an object to the wrong failure state.

## Acceptance criteria

- Identical repeated bindings are idempotent.
- Conflicting `(model, process_name)` bindings fail at startup.
- Bound state fields are validated as concrete model fields.
- For bindings sharing `(model, state_field)`, in-progress ownership is either globally unambiguous or automatic recovery is rejected/disabled with a Django system-check error.
- Tests cover two process classes sharing a model field and in-progress value.

emil-balashov:
Implemented on the PR #137 branch in eda0a0d: identical re-binds are idempotent no-ops; a (model, process_name) already bound to a different process/field raises ImproperlyConfigured at startup (previously the model property was silently overwritten while the registry kept both claims); state_field is validated as a concrete model field at bind time. New django_logic.E001 system check flags an in_progress_state claimed by more than one bound machine on the same (model, state_field), and recover_stranded_states skips such states at runtime with a loud error — instances stay parked and re-drivable instead of getting a guessed owner's failed_state. Closes with #137.

PR #176, opened 2026-07-28
0.10.1 — enforce recovery-ownership instead of in_progress_state uniqueness (#175)

### LINKED ISSUE #175: in_progress_state uniqueness is stricter than the engine needs — enforce recovery-ownership instead
## Problem

`Process.__init_subclass__` calls `_validate_unique_in_progress_states`, which raises `ImproperlyConfigured` when two transitions anywhere in a process tree declare the same `in_progress_state`. That forbids a legitimate and common shape: **several actions on one model that all mean "busy" to a client.**

Concrete case (gv, Borderless360/gv#9039). A `BatchAction` model has three background transitions — `charge_order_estimates`, `create_sales_invoices`, `create_credit_notes` — declared across one process tree, two of them on the same class. All three are "the batch action is running"; the frontend knows exactly one in-progress value, `updating`. The engine refuses the declaration, so the consumer had to:

- invent `updating_charge_order_estimates` / `updating_create_sales_invoices` / `updating_create_credit_notes`,
- add all three to the model's `choices` plus a schema migration,
- add an import-time guard so a new action can't forget its choices entry,
- add a `Process`-level accessor to collect them back,
- and collapse the family onto `updating` in the API serializer.

~30 lines of scaffolding, a migration, and a second source of truth — to work around one assertion. The alternative (drop `in_progress_state` entirely and derive "busy" from an uncompleted `TransitionMessage`) costs a query per row and gives up the durable state.

## The justification is stale

`_validate_unique_in_progress_states`' docstring says:

> Unique `in_progress_state` is what lets the phase-2 background transition lookup work unambiguously — the in-progress state alone identifies the transition that's mid-flight, without any source-state search.

That stopped being true when `owning_process_class` landed on `TransitionMessage` (migration `0007`). Today:

- **phase-2 restore** resolves `(owning process class, action_name)` off the TM row — `_restore` / `_find_background_transition_in_owner` — guarded by `_validate_unique_background_action_names`. It never searches by state.
- **`_state_guard_matches`** compares the persisted state against `transition.in_progress_state` of the transition it has *already* restored. No lookup.
- **`_ensure_no_background_in_flight`** keys on the TM row, not the state.

## The rule that is actually needed

One consumer genuinely needs to know who owns a state: `recover_stranded_states`, for a **record-less** stranded instance (in an in-progress state with no uncompleted TM). It has no provenance, so it picks an owner and drives that transition's `fail_transition`.

Picking is safe exactly when every claimant would recover *identically*. `fail_transition` reads only `failed_state`, `failure_side_effects` and `failure_callbacks` — everything else it touches is logging. So the invariant is:

> Every transition claiming an `in_progress_state` on a given (model, state_field) must agree on `failed_state`, `failure_side_effects` and `failure_callbacks`.

That subsumes the cross-binding case #143 already handles, and it is strictly weaker than name uniqueness. Note this is also why background transitions barely need the sweep at all: `dispatch.recover_stranded_states`' own docstring says their TM row is the durable record the starter/watchdog/finalizer act on, and `_sweep_transition` skips any candidate with an uncompleted row.

## Proposal

1. Delete `_validate_unique_in_progress_states` and its `__init_subclass__` call.
2. Generalize `collect_ambiguous_in_progress_states` to flag a key whose claimants have **more than one distinct recovery signature** `(failed_state, failure_side_effects.commands, failure_callbacks.commands)`, instead of more than one distinct transition object. This is a strict generalization — the existing "same transition object through two bindings is not ambiguous" case has one signature, so #143's behaviour is unchanged.
3. `django_logic.E001` and the sweep's skip then cover both cases with no structural change; only their wording needs to say "recover differently" rather than "claimed by more than one bound machine".

Sharing becomes free where it is safe, and a genuinely ambiguous topology still fails `manage.py check` loudly.

emil-balashov:
Closes #175. Releases **0.10.1**.

## What changed

`in_progress_state` no longer has to be unique. `_validate_unique_in_progress_states` is deleted, and the rule the engine actually needs is the only one enforced:

> Every transition claiming an `in_progress_state` on a given (model, state_field) must recover a record-less stranded instance identically — same `failed_state`, same `failure_side_effects`, same `failure_callbacks`.

`collect_ambiguous_in_progress_states` now keys ambiguity on that recovery signature instead of transition identity. Where claimants agree, sharing is free and the sweep picks any of them; where they disagree, `django_logic.E001` fails `manage.py check` and `recover_stranded_states` skips the state — exactly what it already did across bindings (#143).

## Why the old rule was wrong

Its docstring claimed "the in-progress state alone identifies the transition that's mid-flight, without any source-state search". That has not been true since `owning_process_class` landed on `TransitionMessage` (migration `0007`):

- phase-2 restore resolves `(owning process class, action_name)` off the row (`_restore` → `_find_background_transition_in_owner`), guarded by `_validate_unique_background_action_names`;
- `_state_guard_matches` compares the persisted state against the `in_progress_state` of the transition it has **already** restored;
- `_ensure_no_background_in_flight` keys on the TM row.

The single consumer that needs an owner is `recover_stranded_states`, and only for a **record-less** stranded instance. It drives `fail_transition`, which reads `failed_state` + the two failure bundles and nothing else — everything else it touches is logging. Hence the signature.

## Why this is a strict generalization, not a relaxation of #143

`len({id(transition) …}) > 1` → `len({_recovery_signature(t) …}) > 1`. The same transition object reached through two bindings yields one signature, so #143's "not ambiguous" case is preserved verbatim; and two *different* transitions that disagree still trip. What newly passes is only the case where disagreement is impossible. `_recovery_signature` compares command **identity**, so two equal-but-distinct callables read as different — erring toward calling a key ambiguous.

## What this unblocks

Several actions on one model that all mean "busy" to a client. Borderless360/gv#9039 has three background transitions on `BatchAction` (`charge_order_estimates`, `create_sales_invoices`, `create_credit_notes`) that share one `failed_state` and identical failure hooks, against a frontend that knows exactly one in-progress value. Under the old rule that consumer needed three invented `updating_<action>` states, a schema migration, an import-time guard, a `Process`-level accessor to collect them back, and an API-layer collapse. All of that goes away.

## Tests

`Ran 496 tests … OK` (was 493).

- `tests/background/test_validation.py` — the two tests that pinned the raise now pin acceptance (same class, and across a nested tree).
- `tests/test_binding_validation.py` — three new: shared state + matching recovery is not ambiguous and passes `E001`; shared state + matching recovery **is** swept (a stranded instance reaches `shared_failed`); shared state + divergent `failure_callbacks` is ambiguous and reports `E001` with "recover differently".
- The existing #143 fixtures (`_MachineA`/`_MachineB`, differing `failed_state`) are unchanged and still ambiguous, which is the regression guard for the generalization.

`ruff` error count is identical before and after (58, all pre-existing). `tests.test_metadata` green after the version bump.

## Docs

`E001`'s message + hint, the sweep's skip log, `BackgroundTransition`'s docstring, `Process`'s docstring, `CLAUDE.md` rule 4, the README nested-processes constraints paragraph, and `docs/design/BACKGROUND_TRANSITION_ANALYSIS.md`'s "two smaller rules" bullet.

## Upgrade note

Nothing to do. Every topology that was valid before is still valid; this only stops rejecting ones that were always safe. If you had worked around the raise with per-action in-progress states, you can collapse them — keep the old values in your model's `choices` until no rows hold them.

---

> [!NOTE]
> **Medium Risk**
> Touches stranded-state recovery and `manage.py check` topology rules for background/sync machines; behavior change is a generalization (previously valid configs stay valid) but misconfigured shared states can still park rows until E001 is fixed.
> 
> **Overview**
> **Release 0.10.1** drops class-time **`ImproperlyConfigured`** for duplicate `in_progress_state` within a process tree (`_validate_unique_in_progress_states` removed). Sharing one “busy” value across several transitions is allowed when every claimant on the same `(model, state_field)` would recover a **record-less** stranded instance the same way.
> 
> **`collect_ambiguous_in_progress_states`** now treats ambiguity as disagreeing **recovery signatures** (`failed_state` plus identity of `failure_side_effects` / `failure_callbacks` commands), not duplicate transition objects. **`django_logic.E001`**, **`recover_stranded_states`** skip logging, and docs (`README`, `CLAUDE.md`, `BackgroundTransition` / `Process` docstrings) describe this rule instead of global uniqueness.
> 
> Tests flip from expecting the old raise to accepting shared states with matching recovery, and add coverage for sweep + divergent failure hooks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf931d3296276cd30500da0e02f7cff1da1f2470. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,36 @@
 
 ## [Unreleased]
 
+## [0.10.1] — 2026-07-28
+
+### Changed
+
+- **`in_progress_state` no longer has to be unique.** Declaring two transitions
+  in one process tree with the same `in_progress_state` used to raise
+  `ImproperlyConfigured` at class-creation time. The rule the engine actually
+  needs is narrower, and it is now the only one enforced: every transition
+  claiming an `in_progress_state` on a given (model, state_field) must **recover
+  a record-less stranded instance identically** — same `failed_state`, same
+  `failure_side_effects`, same `failure_callbacks`. Where they agree, sharing is
+  free and `recover_stranded_states` picks any claimant; where they disagree,
+  `django_logic.E001` fails `manage.py check` and the sweep skips the state, as
+  it already did across bindings (#143).
+
+  The old justification — "the in-progress state alone identifies the
+  transition that's mid-flight" — has not been true since 0.6 added
+  `owning_process_class` to `TransitionMessage` (migration 0007): phase-2 restore
+  resolves `(owning process class, action_name)` off the row, guarded by
+  `_validate_unique_background_action_names`, and never searches by state.
+  `_state_guard_matches` compares against the transition it already restored.
+
+  This unblocks the legitimate shape the raise forbade: several actions on one
+  model that all mean "busy" to a client, sharing one in-progress value and one
+  `failed_state`. Consumers previously forced to synthesize a per-action state
+  (and map it back for the API) can collapse it back to one.
+
+  `E001`'s message and hint changed accordingly; nothing else in the public API
+  moved, and a topology that was valid before is still valid.
+
 ## [0.10.0] — 2026-07-27
 
 An overengineering sweep. ~1,900 lines removed across the engine, its tests and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,8 +50,11 @@ change on success) for anything slow, external, or retriable.
    completion check** on the parent, and **aggregate errors by reading child
    rows** (give the parent an explicit `action_required` partial-failure
    state). Never re-raise a child error into the parent.
-4. **`in_progress_state` is unique within a Process**; set a `failed_state` so
-   failures are contained.
+4. **Set a `failed_state`** so failures are contained. Transitions may share an
+   `in_progress_state` (several actions that all mean "busy" to a UI), provided
+   they also share `failed_state` and their failure hooks — otherwise recovery of
+   a record-less stranded instance would guess, and `django_logic.E001` fails
+   `manage.py check`.
 5. **Test in sync mode**: `DJANGO_LOGIC['BACKGROUND_EXECUTION']='sync'` (or the
    `sync_execution()` context manager) runs phase 2 inline with no broker and
    propagates exceptions; `retry_pending()` simulates the periodic starter.
--- a/README.md
+++ b/README.md
@@ -768,7 +768,7 @@ class GmailConversationProcess(Process):
         BackgroundTransition(
             action_name='send_message_via_integration',
             sources=['open'], target='open',
-            in_progress_state='gmail_sending',     # must be unique across the tree
+            in_progress_state='gmail_sending',     # distinct here: different failure hooks
             conditions=[is_gmail],
             side_effects=[send_via_gmail],
         ),
@@ -811,8 +811,11 @@ phase 2 restores that exact transition from the recorded owner — it does not
 re-evaluate the condition, so routing is deterministic even if the instance
 changes mid-flight. Constraints: a background `action_name` must only be
 **unique within a single process class** (two in one class are
-indistinguishable at restore), and every `in_progress_state` must be unique
-across the whole tree. A background `action_name` *may* coincide with a
+indistinguishable at restore). `in_progress_state` may be shared by several
+transitions — useful when a UI only knows one "busy" value — provided they
+agree on `failed_state` and their failure hooks, since a record-less stranded
+instance in that state has no owner to recover under; `django_logic.E001`
+fails `manage.py check` when they disagree. A background `action_name` *may* coincide with a
 synchronous transition of the same name (phase 2 restores only background
 transitions; phase 1 routes the call by condition) — so a synchronous fast-path
 and a durable background slow-path can share one `action_name`.
--- a/django_logic/background/dispatch.py
+++ b/django_logic/background/dispatch.py
@@ -192,18 +192,18 @@ def recover_stranded_states() -> int:
 
     recovered = 0
     seen = set()
-    # In-progress states claimed by more than one bound machine have no
-    # provenance for a record-less stranding — recovering would guess an
-    # owner and could run the wrong failed_state/hooks. Skip them loudly;
-    # the django_logic.E001 system check flags the topology itself (#143).
+    # A record-less stranding has no provenance, so recovery picks one of
+    # the transitions claiming the state. That is only safe while they all
+    # recover identically; where they disagree, skip loudly. The
+    # django_logic.E001 system check flags the topology itself (#143).
     ambiguous = collect_ambiguous_in_progress_states()
     for key in sorted(ambiguous):
         model_label, state_field, in_progress = key
         logger.error(
             f'recover_stranded_states: in_progress_state {in_progress!r} '
-            f'on {model_label}.{state_field} is claimed by multiple bound '
-            f'machines (django_logic.E001); skipping recovery for it — '
-            f'stranded instances stay parked until the binding topology '
+            f'on {model_label}.{state_field} is shared by transitions that '
+            f'recover differently (django_logic.E001); skipping recovery '
+            f'for it — stranded instances stay parked until the topology '
             f'is fixed.'
         )
     for binding, process_cls, transition in iter_bound_transitions():
--- a/django_logic/background/transitions.py
+++ b/django_logic/background/transitions.py
@@ -54,8 +54,9 @@ class BackgroundTransition(Transition):
         - ``in_progress_state`` — if omitted, the state field does not
           change until phase 2 finishes. Providing it is strongly
           recommended so concurrent readers see "in progress" rather
-          than the pre-transition state. ``in_progress_state`` values
-          must be unique within a :class:`django_logic.Process`.
+          than the pre-transition state. Transitions may share one, as
+          long as they also share a ``failed_state`` and failure hooks —
+          see ``django_logic.E001``.
     """
 
     is_background = True
--- a/django_logic/checks.py
+++ b/django_logic/checks.py
@@ -41,12 +41,15 @@ def check_hook_signatures(app_configs, **kwargs):
 
 @checks.register('django_logic')
 def check_unambiguous_in_progress_ownership(app_configs, **kwargs):
-    """Two bound machines must not claim the same in_progress_state on
-    one (model, state_field) — a record-less stranded instance there has
-    no provenance, so automatic recovery could run the wrong transition's
-    failed_state and failure hooks (``django_logic.E001``).
-    ``recover_stranded_states`` also skips such states at runtime, but
-    the topology itself is the defect; fail loudly at check time."""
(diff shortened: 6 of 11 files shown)


emil-balashov:
## Review — implementation matches the plan, two things to decide

Reviewed against the proposal in #175 step by step. All three steps landed as specified: `_validate_unique_in_progress_states` and its `__init_subclass__` call are gone, `collect_ambiguous_in_progress_states` keys on `_recovery_signature` instead of `id(transition)`, and E001 + the sweep log carry the new wording with no structural change. Suite re-run locally: **496 OK**, `tests.test_metadata` 4 OK.

### The load-bearing claims check out

**`_recovery_signature` is complete.** I read `Transition.fail_transition` (transition.py:243-269) to see whether it touches anything the signature omits. It reads `failed_state`, `failure_side_effects`, `failure_callbacks` — and `in_progress_state`, at line 255, for `wrote_state`/`deferrable`. That fourth one is the *map key*, so every claimant shares it by construction and its absence from the tuple is correct, not an oversight. `_release_lock`'s `deferrable` derives from those same two fields. Nothing else.

**The stale-justification argument is true.** `_find_background_transition_in_owner(process, action_name, owning_path)` (runner.py:862) resolves by owning class path + `action_name` and never looks at state, and `_validate_unique_background_action_names` is still enforced in `__init_subclass__` — so the guard that phase-2 restore actually depends on is intact. The remaining `in_progress_state` reads in `background/` are the already-restored comparison (dispatch.py:425) and the sweep's own per-transition candidate query. No lookup-by-state exists.

**#143 is genuinely preserved.** Same transition object through two bindings → identical command `id()`s → one signature. The unchanged `_MachineA`/`_MachineB` fixtures are the right regression guard.

`id()`-not-equality is the correct conservative choice and the docstring says so; all objects are alive for the duration of the comparison, so no GC-reuse hazard.

### 1. The sweep now does N passes over a shared state

`recover_stranded_states`' dedup key still includes `action_name`:

```python
key = (binding.model._meta.label, binding.state_field, in_progress, transition.action_name)
```

Under the old uniqueness rule that was equivalent to deduping per state. Now that sharing is the point, the motivating gv case — three transitions on `updating` — runs `_sweep_transition` three times per sweep, each doing its own pk-keyset paging over the candidate set.

**Not a correctness bug**, and your own test proves it: `test_shared_state_with_matching_recovery_is_swept` asserts `recovered == 1` with three claimants. The `base_qs.filter(**{state_field: in_progress_state})` at dispatch.py:324 means once the first claimant moves an instance to `failed_state`, the siblings match nothing.

But the redundant paging is real, and dropping `action_name` from that key would be safe *precisely because* of the invariant this PR introduces — any claimant recovers identically, so one pass is sufficient by construction. Worth doing here rather than leaving a 3× cost in the feature's headline use case.

One nuance if you keep it as-is: when a candidate's recovery raises, it stays in `in_progress_state`, so a sibling claimant retries it later in the same run instead of the next one. Harmless given identical signatures, just a behaviour change worth knowing about.

### 2. The upgrade note omits a newly-reachable failure mode

"Nothing to do. Every topology that was valid before is still valid" is accurate as far as it goes, but it skips the other direction: a divergent shared `in_progress_state` **within one tree** was previously *impossible* — `__init_subclass__` raised at import. It is now constructible, and the only thing that catches it is E001 at `manage.py check` time.

That matters because **Celery workers don't run system checks.** A consumer who shares a state and mismatches `failed_state` now boots workers cleanly and discovers it as parked rows plus a logged error, where before they got an unmissable `ImproperlyConfigured` at import. E001 failing `check`/`migrate`/`runserver` is a solid mitigation and this is the necessary cost of the feature — but it's a new way to be wrong, and the upgrade note is where someone would look for it.

### 3. Merge order — CHANGELOG conflicts with #174

Both this and #174 insert directly under `## [Unreleased]`, so GitHub reporting both MERGEABLE is misleading; each is clean only against master independently. I tested it: merging this then #174 gives `CONFLICT (content): Merge conflict in CHANGELOG.md`. `README.md`, `CLAUDE.md` and `pyproject.toml` all auto-merge cleanly.

The conflict is trivial (ordering two adjacent sections), but the sequencing decision isn't:

- **Merge this first** → cut and publish 0.10.1, then rebase #174 and cut 0.11.0. Two releases, each with one clear story.
- **Merge both, release once as 0.11.0** → this PR's `[0.10.1] — 2026-07-28` heading and `version = "0.10.1"` need folding into 0.11.0.

Worth picking deliberately, since the version bump and the dated heading are already committed here. Also note the heading is dated **2026-07-28** — if 0.10.1 isn't published today it drifts, which is the same small nit that hit 0.10.0.

emil-balashov:
## Follow-up: `id()`-based signatures false-positive on equivalent hooks

Went back over `_recovery_signature` specifically to ask whether identity comparison is the right primitive. It has a reproducible sharp edge:

```python
def notify(instance, **kw): pass

# same plain function reference in both transitions
a = Transition('a', ..., in_progress_state='busy', failed_state='f', failure_callbacks=[notify])
b = Transition('b', ..., in_progress_state='busy', failed_state='f', failure_callbacks=[notify])
# → signatures equal: True   ✅

# two equivalent partials — what a consumer parameterising a shared hook writes
c = Transition('c', ..., in_progress_state='busy', failed_state='f', failure_callbacks=[partial(notify)])
d = Transition('d', ..., in_progress_state='busy', failed_state='f', failure_callbacks=[partial(notify)])
# → signatures equal: False  ⚠️
```

The common case (a shared module-level function) works, which is why the tests pass. But `partial(...)`, a `lambda`, a bound method, or any hook list built in a helper/factory produces a fresh object per transition, so two transitions that provably recover *identically* get `django_logic.E001`: "shared by transitions that recover differently."

The docstring frames identity as deliberate — "errs toward calling a key ambiguous" — and as a safety default that's defensible. The problem is the **recovery path from a false positive is undiscoverable from the error.** E001's hint says "give them a matching `failed_state` and matching `failure_side_effects`/`failure_callbacks`", which the consumer has *already done*. The actual fix is "hoist the `partial` to a module-level constant so both transitions reference one object" — nothing in the message or the docs points there, and it's a strange thing to have to know.

This is the one spot where I'd argue the design is cleverer than the requirement. Options, cheapest first:

1. **Say so in the hint** — one line: "these must be the *same* callable objects, not merely equivalent ones; hoist shared `partial`/`lambda` hooks to a module-level name." Keeps the conservative semantics, makes the escape hatch findable.
2. **Compare by value instead** — e.g. `(func.__module__, func.__qualname__)` per command, falling back to `id()` for anything without a qualname. More predictable for consumers, slightly weaker (two same-named functions with different behaviour would collide, which is a contrived shape).
3. **Compare `failed_state` only**, and treat the hooks as the consumer's problem. Simplest rule to state, but genuinely weaker — divergent failure hooks would silently run the wrong notifications on a record-less recovery, which is the failure this check exists to prevent. I don't recommend it.

I'd take (1) on its own — it's a one-line change that removes the trap without touching semantics.

Issue #182, opened 2026-07-28
Four definitions the engine accepts but cannot honour: shadowed action_name, shadowed process_name, truthiness-coerced bool setting, unknown settings keys


emil-balashov:
Four related gaps where the engine accepts a definition it cannot honour. All reproduced.

**1. An `action_name` colliding with a real `Process` attribute is silently unreachable.** `Process.__getattr__` is only consulted when normal lookup *fails*, so a transition named `state`, `is_valid`, `transitions`, `process_name`… is shadowed by the attribute. `get_available_actions()` still advertises it, so it looks live and does nothing:

```
available: ['is_valid', 'state']
inv.shadow_process.is_valid() returned True   # the bound method, not the transition
```

Nothing rejects it at class or bind time. **Fix:** reject in `__init_subclass__` — any `action_name` that is an existing `Process` attribute is a definition error.

**2. `bind_model_process` accepts a `process_name` that shadows a concrete model field.** It validates that `state_field` exists and is concrete, and that `process_name` is not already bound, but never checks `process_name` against the model's own field names before `setattr(model, name, property(...))`. The property replaces the field's `DeferredAttribute`, so **the model can no longer be instantiated at all**. **Fix:** reject at bind time, alongside the existing checks.

**3. `STRICT_KWARGS_SERIALIZATION` is truthiness-coerced.** `return bool(_conf().get('STRICT_KWARGS_SERIALIZATION', False))` — any non-empty string enables it, and unlike the six other accessors it is never validated by `validate_on_ready`:

```
False   -> strict=False  (warn + drop)
'false' -> strict=True   RAISED KwargsSerializationError
```

Reading `DL_STRICT=false` from an env var silently turns strict mode on. **Fix:** validate it as a real bool at boot, like the numeric settings.

**4. No unknown-key rejection.** A typo'd `DJANGO_LOGIC` key is silently ignored; `W003` only knows the five keys 0.10.0 removed. **Fix:** generalise `W003` — the union of known and removed keys is closed, so anything else is a typo worth reporting.

emil-balashov:
Fixed in PR #183 (`release/0.12.0`) — in the PR's `Closes` list; closes on merge.

[...]

Issue #181, opened 2026-07-28
Rendering {{ obj.process.action }} in a Django template executes the transition (missing alters_data)


emil-balashov:
The closure returned by `Process.__getattr__` (process.py:109) carries no `alters_data` / `do_not_call_in_templates` marker, and Django's `Variable._resolve_lookup` calls any callable it resolves. So a template that *references* a transition instead of listing `get_available_actions` **drives the state machine during rendering**.

Reproduced:

```python
inv = Invoice.objects.create(status='draft')
Engine().from_string('[{{ inv.tpl_proc.approve }}]').render(Context({'inv': inv}))
```

```
status before render: draft
template output     : [c4c6982a-f266-49c1-83f4-e4e815fc9e87]
status AFTER render : approved      <-- the transition ran
```

This is exactly what Django's `alters_data` exists to prevent — `Model.delete` and `Model.save` both set it. A template typo, or an autogenerated admin/debug template that walks attributes, silently mutates state. The rendered output is the `tr_id`, which also leaks an internal identifier into the page.

**Fix:** set `alters_data = True` on the closure in `Process.__getattr__` (one line). Django then renders `''` instead of calling it. Worth doing for `Process.__getattr__`'s returned callable in every form it takes.

emil-balashov:
Fixed in PR #183 (`release/0.12.0`) — in the PR's `Closes` list; closes on merge.

`alters_data = True` on the `__getattr__` dispatcher (the only template-reachable mutating callable — underscore names are refused, and read-only methods stay template-callable by design). The docs reviewer executed the pin: `{{ order.order_process.deliver }}` renders `''` and the state is unchanged after render.

emil-balashov:
Fixed in **0.12.0** — released as [v0.12.0](https://github.com/Borderless360/django-logic/releases/tag/v0.12.0) ([PyPI](https://pypi.org/project/django-logic/0.12.0/)), merged in #183 (`cc046e8`).

Validated across five review passes, the full CI matrix (Python 3.11–3.14 × Django 4.2–6.0, plus both PostgreSQL stability jobs), and three Heroku rig gates on real broker/PostgreSQL/worker infrastructure (31 matrix rows, 0 failures on the final candidate). Every fix is pinned by a test verified to fail when its fix is reverted.

Issue #180, opened 2026-07-28
A nested Process reachable by two paths makes its transitions uncallable (no visited-set in the runtime tree walk)


emil-balashov:
`_iter_available_with_owner` (process.py:237) recurses into `self.nested_processes` with **no visited-class set**, so a nested Process reached by two paths yields each of its transitions twice. `_resolve_transition_with_owner` then sees `len(matches) == 2` and raises `TransitionNotAllowed("There are several transitions available")` — for a *single* declaration.

It is the only one of the library's five tree walkers without dedupe: `_iter_process_tree` (process.py:290) guards with `_seen`, `collect_ambiguous_in_progress_states` (process.py:517) with `seen_cls`, `coverage.iter_bound_transitions` (coverage.py:87) with `seen`.

Reproduced with a plain diamond — `RefundProcess` nested under both `PaidProcess` and `ShippedProcess`, both nested under the bound root:

```
get_available_actions()     : ['refund']
get_available_transitions() : ['refund', 'refund']
calling refund(): TransitionNotAllowed: There are several transitions available
```

Note the inconsistency: `get_available_actions()` set-dedupes and **advertises** `refund`, so a UI renders the button and every click fails. The logged hint ("specify conditions and permissions to disambiguate") is unactionable — both matches are the same `Transition` object with the same owning class, so no condition can separate them. `nested_processes = [RefundProcess, RefundProcess]` (a copy-paste) fails identically, and `NextTransition.execute` hits its own `len(transitions) > 1` branch and silently drops the chain.

The same missing guard makes a nested **cycle** recurse forever: with `A.nested_processes = [B]` and `B.nested_processes = [A]`, `get_available_actions()` dies with `RecursionError`. Nothing rejects that topology at class or bind time — `__init_subclass__` only calls the cycle-safe `_iter_process_tree`.

README's "Nested Processes — build complex workflows by combining processes" documents no restriction against reusing a sub-process.

**Fix:** thread a visited-class set through `_iter_available_with_owner`, mirroring `_iter_process_tree`. Deduping by class preserves the supported condition-disambiguated pattern (the same `action_name` on *distinct* nested classes still yields two matches). The five near-identical tree walks should share one helper.

emil-balashov:
Fixed in PR #183 (`release/0.12.0`) — in the PR's `Closes` list; closes on merge.

[...]

Issue #179, opened 2026-07-28
Timeout watchdog is blind to live attempts and re-charges the same one every tick (timeout= makes a transition less reliable)


emil-balashov:
`tm.mark_as_started()` (runner.py:485) runs **inside** `_run_atomic`'s `transaction.atomic()`, so `started_at` is invisible to every other connection until the attempt commits. Two consequences.

**1. The watchdog can never see the attempts it exists for.** `_watchdog_stale_attempts_inline` filters `started_at__isnull=False`, which excludes every genuinely in-flight row. A hard-killed attempt rolls `mark_as_started` back with everything else. So the only rows it *can* act on are rows whose last attempt already committed a retryable failure — rows that already charged themselves one error.

**2. It then charges that same attempt again on every tick,** because `abandon_timed_out_attempt` never clears or advances `started_at`. Organic repro — a real `BackgroundTransition(timeout=1)` whose side-effect sleeps 1.2s then raises, sync mode, MAX_ERRORS=5:

```
after 1 REAL attempt : errors=1  last_error='slow boom'
watchdog tick 1: touched=1 errors=2 last_error='[watchdog timeout] attempt exceeded timeout_seconds=1'
watchdog tick 2: touched=1 errors=3
watchdog tick 3: touched=1 errors=4
```

Three ticks took `errors_count` 1 → 4 with **zero new attempts**; a fourth terminalises the row at MAX_ERRORS and writes `failed_state`, discarding every remaining retry. The real cause (`'slow boom'`) is also overwritten by the synthetic timeout message, so the operator loses it.

In production (watchdog 120s, retry 60s) every attempt that outruns its `timeout` is charged at least twice — so **declaring `timeout=` makes a transition strictly less reliable than omitting it**.

No existing test catches this: every case in `tests/background/test_watchdog.py` hand-crafts a row with an artificial `started_at`, a state the real lifecycle cannot produce while an attempt is live.

**Same root cause** breaks the retry starter's recency guard (`_retry_pending_inline`, tasks.py:116): `Q(started_at__isnull=True) | Q(started_at__lt=cutoff)` cannot skip a live attempt, and a never-attempted row has `started_at IS NULL` forever, so it is re-dispatched on every tick — collapsing the intended `RETRY_MINUTES × MAX_ERRORS` window.

[...]

PR #183, opened 2026-07-29
0.12.0 — correctness release: five engine defects, five review passes, and in_progress_state made background-only

### LINKED ISSUE #178: A failing state write escapes phase-2 error accounting: row retried forever, side-effects re-run forever
`_run_atomic` savepoints the side-effect loop precisely so "the error bookkeeping below always works" (its own comment). But the two `set_state` calls are outside any savepoint and any handler: `_handle_success` (runner.py:536) and `_handle_failure`'s terminal branch (runner.py:582). If either raises — CHECK constraint, `pre_save` receiver, `save()` override, column length — the exception escapes the atomic, **rolling back `mark_as_started`, `record_error` and `mark_as_completed` with it**.

Reproduced with a `pre_save` receiver that rejects the target value (sync mode, MAX_ERRORS=5, RETRY_MINUTES=0):

```
after attempt 1 : errors=0 completed=False status='running' side_effect_calls=1
after 4 retries : errors=0 completed=False status='running' side_effect_calls=5
detect_stuck finalized: 0   (needs errors>=MAX_ERRORS — never reached)
FINAL           : errors=0 completed=False  -> retried forever
```

`errors_count` stays 0 forever, so `retry_stale_transitions` re-dispatches every tick and the side-effects — including non-idempotent external calls — re-run indefinitely. `detect_stuck_transitions` never sees it (errors < MAX_ERRORS); `recover_stranded_states` skips it (an uncompleted row exists). The instance is stuck in its `in_progress_state` permanently.

The terminal branch is worse: it rolls back the `record_error` immediately preceding it, pinning `errors_count` one below MAX_ERRORS forever. And `_finalize_terminal_from_watchdog` (runner.py:344) has the identical unguarded write, so the safety net cannot rescue it either.

**Fix:** the target write belongs *inside* the attempt savepoint (it is part of the attempt, and this also restores the documented all-or-nothing-per-attempt contract). The `failed_state` writes need their own savepoint with the failure recorded and `mark_as_completed` still reached, so the retry loop always terminates.

emil-balashov:
A correctness release. No new capability beyond the checks that make two of the defects impossible to reintroduce silently. Closes #178, #179, #180, #181, #182, #154, #188.

**+4,945 / −1,904 across 55 files.** 499 → **568** tests (600 at the pass-4 peak; the marker removal then retired the stranded-sweep and E001 suites wholesale).

> **Read the fourth-pass section at the bottom before the rest.** It found two
> more defects that could lose work, an undeclared breaking change, and nine
> fixes below that had no pinning test at all — so parts of the narrative above
> were true of the code but not yet true of the test suite. Everything is fixed
> and pinned now, and the Heroku gate has run.

## How this was produced, because it matters for how you read it

Three adversarial review passes, each one reviewing the previous pass's diff. Every defect was reproduced before being fixed and re-run after.

| pass | reviewed | confirmed | of those, regressions the previous pass introduced |
|---|---|---|---|
| 1 | 0.11.0 | 20 | — |
| 2 | pass 1's fixes | 18 | **12** |
| 3 | pass 2's fixes | 13 | **5** |

The regression rate is falling but has not reached zero. Pass 3's headline finding was **a test I had written one commit earlier that proved nothing** — it asserted an `isinstance()` on a non-MTI model, so it passed with the fix removed. That is the failure mode worth keeping in mind while reading "528 tests pass" below.

## The five original defects (#178–#182)

**#178 — a failing state write escaped phase-2 error accounting.** The side-effect loop was savepointed so "the error bookkeeping below always works", but both `set_state` calls sat outside any savepoint. A write the database rejected — CHECK constraint, `pre_save` receiver, `save()` override, column length — escaped the outer atomic and **took `record_error` with it**:

```
before: errors=0 after 4 retries, side_effect_calls=5 → retried forever
after : errors=1 → retries charge → errors=3, completed, terminalised
```

No safety net could rescue it: `detect_stuck` needs `errors >= MAX_ERRORS`, `recover_stranded_states` skips rows with an uncompleted message, and the safety-net finaliser had the identical unguarded write. The target write moved inside the attempt savepoint (which also restores the documented all-or-nothing contract); the `failed_state` writes are savepointed and can no longer prevent completion, because completing the row is what stops the loop.

**#179 — the timeout watchdog was blind, and charged the same attempt forever.** `mark_as_started` ran inside the attempt's atomic, so `started_at` was invisible while the attempt ran and rolled back when a worker died — it could never see the attempts it exists for. The only rows it matched had already charged themselves an error, and it re-charged them every tick:

```
before: 1 real attempt → errors 1; three ticks, no new work → 4
after : recorded failure → touched=0, errors stays 1, real cause preserved
        crashed attempt  → touched=1 once, then 0
```

One more tick and the row terminalised with four of its five retries spent on a bookkeeping bug. **Declaring `timeout=` made a transition strictly less reliable than omitting it.**

**#180 — a nested `Process` reachable by two paths was uncallable.** `get_available_actions()` set-deduped and advertised the action; every call raised `TransitionNotAllowed: several transitions available`, with a hint no condition could satisfy. A nested cycle recursed to `RecursionError`.

**#181 — rendering `{{ obj.process.action }}` executed the transition.** Django calls any callable a template resolves. `alters_data` now says no.

**#182 — four definitions the engine accepted but could not honour**: a shadowed `action_name` (advertised, silently did nothing), a `process_name` that made the model uninstantiable, `bool()`-coerced strict flags where the string `'false'` turned strict mode **on**, and no unknown-key reporting.

## What pass 2 and 3 found in those fixes

Worth calling out because the same class of bug is what CI missed:

- **A bare `UPDATE` placed before `select_for_update(nowait=True)`** defeated the phase-2 row-lock contract entirely — a duplicate dispatch waited out the live attempt and re-ran it. Caught by `tests/background/test_concurrency_pg.py`, which **no CI job or make target had ever run**: master OK in 0.093s, the branch FAILED in 10.1s. That file is now in `stability.yml`, and it was the only unrun one.
- **The stranded sweep started reporting recoveries that never happened** — #178 made `fail_transition` swallow a rejected write, so "returned without raising" stopped meaning "the state moved".
- **The deferred-unlock fix broke the mechanism it protected**: `_run_in_savepoint` tracks entries by index window, and an eager `registry.clear()` could drop unlocks an enclosing window owed a release.
- `ScenarioAssertions` read `self.process_name` in four places, which a `None` default turned into `process_name=None` — assertions passing or failing for the wrong reason.
- **A NUL byte in an exception message made `record_error` itself fail** on PostgreSQL, rolling back the accounting with it. Same shape as #178, reached through the bookkeeping. A shared `db_safe_text()` now guarantees the bookkeeping is never the statement that fails.
- The bind-time collision check was wrong in **both** directions, then wrong again: it flagged reverse-FK query names, missed non-field descriptors, and after being rewritten rejected legitimate multi-table inheritance.

## Verification (original pass — superseded by the fourth-pass section below)

| | |
|---|---|
| SQLite | **528 OK** (now 600) |
| real Redis (`tests.settings_redis`) | **528 OK** |
| Postgres + Redis stability | **52 OK** |
| PG concurrency (newly wired into CI) | **2 OK** |
| metadata drift | **4 OK** |
| `manage.py check` | clean |
| `--reverse` / `--shuffle` | both OK |
| `makemigrations --check` | no changes |
| README Quick Start, Complete Example, all 17 documented imports | re-run after every edit |

## Documentation

The **Complete Example is runnable as printed** — its conditions called `instance.items.all()` and read `instance.shipping_address` on a model declaring neither, so a verbatim copy raised an uncaught `AttributeError` (a 500, since the example's own view catches only `TransitionNotAllowed`), and the Troubleshooting section's recommended diagnostic raised the same thing. "Custom State Classes" now names `state_class`, without which the recipe it shows does nothing. Also corrected: the `started_at` description, the claim that 0.8/0.9.0 coverage logs are still read, six stale statements in the design doc `CLAUDE.md` tells engine-changers to read first, and an anchor collision sending three links to the wrong section.

## Removed

The `drf` extra (nothing has referenced `rest_framework` since 0.4), the duplicate failure-callback runner, `_validated_number`'s unreachable `allow_zero`, a defensive branch its own comment called unreachable, and four of the five copies of the nested-process tree walk — not cosmetic, since the un-deduped fifth **was** #180.

## Before merging — two things I would not skip

1. **A Heroku rig pass.** #179 and #178 change behaviour under real worker crashes and rejected writes, which the local suite can only approximate. The rig has cells for this.
2. **The version is deliberately breaking-flagged.** See `Changed (breaking)` in the CHANGELOG: three declarations now raise at import/bind time, the two strict flags now reject non-bools, and `W004` is new (a project keeping unrelated `DJANGO_LOGIC` keys and running `check --fail-level WARNING` needs `SILENCED_SYSTEM_CHECKS`). Every one of those was already broken in a project that hits it — but it will now fail loudly rather than silently.

**16 low-severity findings remain unactioned** and are listed in the review notes rather than fixed here: a CHANGELOG note for the coverage-key shape change, `snapshot()` omitting three fields, the append-truncation dropping the newest note, and `_condition_fingerprint` still not separating callable-instance conditions by configuration.

---

> [!NOTE]
> **High Risk**
> Changes core state-machine durability (background phase-2, watchdog, locks, breaking removal of stranded recovery) and several breaking bind/settings contracts; upgrades need migration for sync `in_progress_state` and pre-upgrade cleanup of legacy stranded rows.
> 
> **Overview**
> **0.12.0** is a correctness-focused release: phase-2 accounting, timeouts, locking, and restore paths are hardened; several misconfigurations now fail at bind/boot; observability improves for lock races and expected contention.
> 
> The largest **breaking** change is architectural: **`in_progress_state` is background-only** — sync `Transition`/`Action` declarations that used it now raise at class creation. **`recover_stranded_states`** (Celery task, dispatch sweep, beat entry) and **`django_logic.E001`** are removed; **`beat_schedule()`** documents **four** safety-net tasks. Killed sync work rolls back to source and becomes re-drivable after lock TTL; README/CHANGELOG describe the explicit busy-state + `next_transition` + periodic re-drive migration.
> 
> **Engine fixes** include moving target/`failed_state` writes inside savepoints (#178), committing **`started_at`** via **`stamp_attempt_started`** before the attempt atomic with **`select_for_update(nowait)`** (#179), watchdog staleness re-checked under row lock with one-charge-per-attempt guards, attempt savepoints via **`_run_in_savepoint`** with **`require_commit`** for suppressed DB errors, safety-net finalizers honoring manual fixes with **`[superseded]`**, and **`db_safe_text`** / kwargs rules (NUL, surrogates, reserved **`user_id`**). Sync path gains **`SourceStateChanged`**, template **`alters_data`**, lock-failure logging (#188), and deferred-unlock registry fixes.
> 
> **Checks & packaging**: new **`django_logic.E003`** (background app missing), **`django_logic.W004`** (unknown `DJANGO_LOGIC` keys), stricter bool settings, **`__all__`** on the top-level package, **`[drf]` extra removed**, coverage keys improved for partials/callable conditions.
> 
> **CI**: **stability.yml** now runs **`tests.background.test_concurrency_pg`** against real PostgreSQL so the phase-2 row-lock contract cannot regress unnoticed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd1445a876c0a14162d8b30b6d7d929bfcedd2a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

# Fourth review pass (commit `3a92708`) — and the Heroku gate

The "two things I would not skip" above are done: a Heroku rig pass, and a
review of the fixes themselves. The review came first, because it changed what
needed validating.

**Seven independent readings of the library** (background engine, locking and
savepoints, sync engine, checks/settings/serializers, testing framework and
coverage, docs/packaging/CI truthfulness) **plus mutation testing of every test
this branch added.** Pass 3's headline was a test that proved nothing; pass 4
started from the assumption there were more, and there were.

| pass | reviewed | confirmed | of those, regressions the previous pass introduced |
|---|---|---|---|
| 1 | 0.11.0 | 20 | — |
| 2 | pass 1's fixes | 18 | **12** |
| 3 | pass 2's fixes | 13 | **5** |
| **4** | **pass 3's fixes + the whole library** | **~30** | **2 serious, both reachable only after this branch's changes** |

## The two that could lose work

**The timeout watchdog could charge — and terminalise — a perfectly healthy attempt.**
`#179` above made `started_at` visible and added a one-charge-per-attempt guard.
But the watchdog still decided *"this attempt is stale"* in its unsynchronised
candidate scan and never re-verified under the row lock. A retry that stamps a
fresh `started_at` in the scan→lock window defeats the new guard precisely
because the guard compares the last error to `started_at`: the new stamp is
*later* than the previous attempt's error, so "no error since this attempt
started" reads true. Reproduced: an attempt **0.00s old** charged a synthetic
`TimeoutError`, and at `MAX_ERRORS` the row finalized to `failed_state` while
its worker was still running. `beat_schedule()` co-schedules the starter at 60s
and the watchdog at 120s on one queue, so this is systematic, not exotic.
The staleness verdict is now re-checked on the locked read.

**A rolled-back phase-2 attempt leaked the state lock of any instance its side-effects had driven.**
Under `DEFER_UNLOCK_UNTIL_COMMIT`, a side-effect that synchronously drives
another instance — what the fan-out and report-back recipes tell you to do —
registers that instance's unlock on `transaction.on_commit` *inside the attempt
savepoint*. A later side-effect failure makes Django discard the hook with the
savepoint while the outer transaction still commits the bookkeeping: the other
instance's state write rolls back but its lock is held **until `LOCK_TIMEOUT`
(7200s)**, every transition on it raises `TransitionNotAllowed("State is
locked")`, and the driver's own retries then burn `MAX_ERRORS` against that held
lock. Every hook bundle already ran through the machinery built for exactly this
(`_run_in_savepoint`) — the attempt savepoint was the one raw `atomic` left, and
**moving #178's target write inside it added a new way to reach the leak.**

## What mutation testing found

27 mutations against the tests this branch added. Two assertions proved nothing,
and **nine fixes had no pinning test anywhere** — their reverts survived the
entire 528-test suite:

- the `'Set state'` vs `'Set State'` filter (Cursor Bugbot's finding, confirmed with a passing mutation);
- `test_target_write_rolls_back_its_own_attempt`, whose transition had only a `_noop` side-effect, so there was nothing to roll back;
- unpinned: **#178's terminal path** (the headline), the watchdog finalizer's savepointed write, `Action.fail_transition`'s, `db_safe_text`, `ScenarioAssertions._process_name` (named in this PR's own pass-2 list), `State.set_state`'s attribute restore, `_run_in_savepoint`'s silent-rollback branch, `get_persisted_state`'s routing, and `STRICT_HOOK_SIGNATURES`' `is True`.

## Also fixed

Safety-net finalizers now honour a manual fix **whole** — they used to skip only
the `failed_state` write and still run `failure_side_effects` and
`failure_callbacks` against an instance an operator had already resolved.
Unclassified `_restore` failures are accounted instead of escaping with
`errors_count=0` (the #178 infinite-retry class, reached through a different
door). The stop-retry write routes through `db_safe_text`. The two phase-2 tree
walks get #180's cycle guard. `failed_state == in_progress_state` and
lock-leaking engine-parameter kwargs (`exception=`, `deferrable=`) are refused.
`E003` reports a missing background app. `E001` no longer crashes on duck-typed
transitions (which took `manage.py check` **and every beat tick of the stranded
sweep** down with it) and no longer misses MTI/proxy column sharing. The
shadowed-`action_name` check stops rejecting working topologies. **#154** is
closed. On the testing side: vacuous bare-string assertions, lossy snapshot
round-trips, and coverage keys that collapsed configured condition instances
into one key — greenlighting untested transitions, the per-courier pattern the
fingerprint exists for.

## Undeclared breaking changes, now declared

The docs reviewer found **five behaviour changes a consumer can hit that the
CHANGELOG did not mention** — the highest-severity documentation finding, since
this release is deliberately breaking-flagged. Chief among them: NUL bytes and
lone surrogates in background kwargs now raise `KwargsSerializationError` at
phase 1, on every backend, regardless of strict mode. All five are in `Changed
(breaking)` now, along with `State.get_persisted_state()`'s alias routing, two
`TransitionMessage` column shapes, and the coverage-key change.

## Verification

**Local — 600 tests** (528 → 600), green **forward, `--reverse`, and `--shuffle`**.
`manage.py check` clean, `makemigrations --check` clean, metadata drift green.
**Every fix is pinned by a test verified to fail when its fix is reverted** —
10 mutation checks, all caught. The intermittent `BooleanSettingTests` /
`UnknownSettingsCheckTests` flakes two reviewers reported were traced to the
mutation-testing agent editing those very files concurrently; both are stable
across the ordered runs above.

**Heroku — `django-logic-test`, release v78, engine `3a92708` verified on the dyno:**

| group | rows | result |
|---|---|---|
| preflight | broker, DB, cache, `STARTER_QUEUE`, dyno topology | **OK** |
| P0 | TC-01, TC-04 ⭐ crash, TC-18 ×2, TC-13, TC-12 ×2, TC-16 | **9 / 9** |
| **012 (new)** | TC-29a/b/b-ii/c (#178), TC-30a ×2/b ×2/c/d (#179) | **11 / 11** |
| P1 safety nets | TC-09 (rewritten), TC-10, TC-15, TC-11 | **5 / 5** |
| concurrency (real PG) | TC-07, TC-08 | **3 / 3** |
| on-dyno suite | full pytest on a `heroku run` dyno | **85 passed** |

The rig gained **TC-29** and **TC-30** for the two fixes with real-infrastructure
surface, and **TC-09 was rewritten** — its old "tick the watchdog 5× on one
stamp" recipe is wrong by design now that an attempt is charged at most once.

What only real infrastructure could prove: **TC-29c**, a NUL byte in an
exception message (SQLite stores it happily; PostgreSQL rejects the write, which
is how the accounting write itself became the failing statement); **TC-30c**, a
live attempt holding a genuine `select_for_update(nowait)` row lock, which the
watchdog correctly defers to; **TC-07/TC-08**, both needing two real
connections; **TC-04**, a `SIGKILL`ed worker redelivered by `acks_late`.
Observed live: `watchdog_stale_attempts` firing on beat's own schedule and
returning 0 while nothing is stale — the answer the pre-fix engine could not
give. Every group passed its `fleet clean` assertion. Dynos parked afterwards.

Harness work is on `Borderless360/django-logic-test@chore/validate-0.12.0`
(matrix rows, `veto_states` / `nul_error` injection, a re-entrant `simulate
timeout`, and `docs/RESULTS.md`).

## Added after the Heroku gate

**#188** (Andrey, from a live gv incident 2026-07-29): a failed `state.lock()`
raised before logging anything, so a frozen instance was indistinguishable from
a healthy start — seven conversations re-driven for ten days produced ~1400
`Start` lines and zero indication a leaked lock was the cause. Fixed in
`253c080`: both call sites log `Lock failed <instance_key> — state is locked`
at INFO (per #154's position) before raising, the phase-1
conditions/permissions rejection logs too, and the `Lock`/`Unlock` lifecycle
lines now carry `instance_key` so a per-instance log filter shows the whole
lock lifecycle without a `tr_id` self-join. Pinned by
`tests/test_lock_logging.py` (mutation-checked), `docs/logger.md` updated.
Log-format note: `Lock`/`Unlock` gained a trailing `instance_key`. This is
observability only — no lock behaviour changed — so the Heroku evidence above
stands for `253c080`.

## The design cut: `in_progress_state` is background-only; the sweep and E001 are retired

Decided with Andy after the fourth pass, executed in `dfe3a1a`: **the
in-progress marker now exists only where it has a durable recovery owner.** On
a `BackgroundTransition` it is written atomically with the `TransitionMessage`
row, so the TM safety nets own the whole flight. On a synchronous transition it
was a marker with no record — a hard-killed worker parked the instance in a
state with no outbound edges (#136), and the engine grew its most defect-dense
subsystem (the stranded sweep and the E001 ownership rules; the majority of
serious findings across all four review passes) just to cope with the marker
it had itself written. Declaring it on a sync `Transition`/`Action` now raises
at class creation; a killed sync run rolls back to its source state and is
simply re-drivable.

Deleted without regret: `recover_stranded_states` (fifth beat task and ~350
lines of sweep — record-less stranding is now structurally impossible),
`django_logic.E001` and the whole recovery-ownership cluster, the sync
marker-write, and the marker half of the DEFER-unlock rules. `beat_schedule()`
ships **four** entries. The replacement for sync consumers — a real busy state
entered by a fast transition, chained via `next_transition`, with a three-line
periodic re-drive for the crash window — is documented in the README and
pinned end-to-end in `tests/test_explicit_busy_state.py`. The re-drive retries
*forward*, which the sweep's force-fail never could; the one behaviour an
operator will notice is that rows completed without a managed `failed_state`
write now leave the instance parked-but-re-drivable instead of force-failed on
a later sweep tick. **Instances already stranded by a previous release must be
re-driven or cleaned up before upgrading** — the CHANGELOG migration note has
the details.

Verification of the cut: 565 engine tests green forward/`--reverse`/`--shuffle`
(the background-only gate mutation-checked), CI green across the matrix, and a
full second Heroku gate at `dfe3a1a` (releases v80/v81): **31 rows, 0
failures** — P0 9/9, 012 11/11, P1 5/5, concurrency 3/3, durability 3/3
(riding through a mid-run deploy), 78 on-dyno tests. TC-24 and TC-25 are
retired *by design*: the stranding they reproduced cannot be expressed
anymore. TC-29b's end state is now pinned as a forward re-drive.

## Still open, deliberately

Filed rather than fixed here, none of them regressions of this branch:
`State.instance_key` omits the DB alias (cross-tenant lock contention);
`Action.fail_transition`'s `is_locked`→write TOCTOU; the full MTI/proxy lock-key
rework (changing the key shape is a rolling-deploy hazard, so E001 detection
lands now and the key does not); and `get_persisted_state`'s routing needs a
second DB alias in the test settings to pin.

---

# Fifth pass — reviewing the branch for release, and the third Heroku gate

Two things the passes above had not covered: Cursor Bugbot's last review was
`253c080`, so **`dfe3a1a` — the 348-line design cut, the largest behavioural
change in this PR — had been reviewed by nothing but CI and the rig**; and the
one High-severity Bugbot finding on the branch was still open.

## The finding, reproduced and narrowed

**A phase-2 attempt whose writes were all discarded was reported as a success.**
Django's `mark_for_rollback_on_error` (which `Model.save_base` wraps every write
in) flags the connection when a database error is raised inside an atomic block
*even if the caller catches it* — the
`try: obj.save() except IntegrityError: pass` idiom without the nested
`atomic()` it needs. `Atomic.__exit__` then discards the savepoint with **no
exception propagating**. `_run_in_savepoint` already detected that case (it
releases the deferred unlocks the rollback dropped) but returned normally, and
phase 2 reads "returned" as "committed".

Bugbot's mechanism is right; its conclusion did not reproduce. A/B against
`master`:

| | 0.11.0 (master) | this branch, before the fix | after |
|---|---|---|---|
| `BackgroundAction` | completed, `errors=0`, callbacks ran, writes gone | **identical** | accounted, retried, terminal |
| `BackgroundTransition` | **completed and advanced to `target`** | `TransactionManagementError` → accounted | unchanged |

So the reachable case is `BackgroundAction`, which writes no state — a
`BackgroundTransition` is protected *by accident*, its target `set_state` being
the last statement in the attempt and raising on the poisoned connection.
**Pre-existing, not a regression of this branch** (which strictly improves the
transition case). Fixed anyway, because this release's own thesis is that a
rolled-back attempt is never a success. Correctly-written consumer code — the
suppression inside its own `atomic()` — is unaffected and is pinned as such.
The `require_commit` pin is mutation-checked: reverting the flag fails both
pins and leaves the control test green.

## Eight statements the design cut retired but left shipping

The prose and comments moved with `dfe3a1a`; the **runtime strings did not.**

- the deployment section promised `beat_schedule()` "routes all **five** tasks …
  stranded 300s", naming a `stranded_seconds=` keyword that now raises
  `TypeError` — and the Safety-net section said "**Five** periodic tasks" above
  a list of four;
- `BackgroundTransition`'s own docstring still made a shared `in_progress_state`
  conditional on matching failure hooks and cited the deleted
  `django_logic.E001`, which the README and `CLAUDE.md` correctly call retired;
- **three operator-facing log lines** sent whoever read them to
  `recover_stranded_states`, deleted in the same commit;
- `LOCK_TIMEOUT`'s and `failed_state == in_progress_state`'s
  `ImproperlyConfigured` messages still justified themselves by stranded
  recovery;
- and one docstring sentence was left ungrammatical by the edit.

Verified sound while reading the cut: the class-creation gate (`is_background`
is a class attribute, so the ordering holds), `BackgroundAction` still refusing
the marker, and the claim the whole replacement rests on — `in_progress_state`
really is appended to `sources` (`transition.py`), so a parked instance is
re-drivable.

## Verification

**Local — 568 tests** (565 + 3), green forward, `--reverse`, `--shuffle`, and
against **real Redis** (`tests.settings_redis`). Metadata drift green,
`manage.py check` clean, `makemigrations --check` clean. **CI green across the
full matrix** (Python 3.11–3.14 × Django 4.2–6.0, `django-main`, both
PostgreSQL stability jobs). Bugbot re-reviewed `bd1445a` — which carries both
the design cut and this fix — and **found nothing**.

**Heroku — `django-logic-test`, release v82, engine `bd1445a` verified on the
dyno** (`_SilentRollback` present, `require_commit` in the signature, four beat
entries, sweep gone):

| group | rows | result |
|---|---|---|
| preflight | broker, DB, cache, `STARTER_QUEUE`, dyno topology | **OK** |
| smoke | healthz, orders.fulfill | **3 / 3** |
| P0 | TC-01, TC-04 ⭐ crash, TC-18 ×2, TC-13 ⭐ action, TC-12 ×2, TC-16 | **9 / 9** |
| 012 | TC-29a/b/b-ii/c (#178), TC-30a ×2/b ×2/c/d (#179) | **11 / 11** |
| concurrency (real PG) | TC-07, TC-08 | **3 / 3** |
| P1 safety nets | TC-09, TC-10, TC-15, TC-11 | **5 / 5** |
| durability (real beat) | TC-02, TC-03 | **3 / 3** |
| on-dyno suite | full pytest on a `heroku run` dyno | **78 passed** |

**31 matrix rows, 0 failures**, every group passing its `fleet clean`
assertion — the same coverage as the second gate, re-run because this fix
changes the phase-2 attempt path. TC-13 (`BackgroundAction` happy path) and
TC-29c (a NUL byte in an exception message on real PostgreSQL) are the rows
that speak directly to it. Dynos parked at 0 afterwards.

The final head `c8dc519` differs from the gated `bd1445a` by one README line
and one `ImproperlyConfigured` string on a branch a valid `LOCK_TIMEOUT` never
enters, so the gate above stands for it.

## Consumer checks (`RELEASING.md` step 6)

Both known downstreams are compatible with the breaking changes, checked by AST
scan rather than grep:

- **gv** — **0** sync `in_progress_state` declarations and **0** bare-string
  `sources`, so nothing raises at import or bind time; every
  `DJANGO_LOGIC_NEW` key is inside `_KNOWN_SETTINGS`, so no `W004`; and beat is
  installed programmatically through `beat_schedule()`, so it drops to four
  entries on re-vendor with no edit.
- **the rig** — `chore/validate-0.12.0` already retired the fifth beat entry
  and `run_starter --stranded`. ⚠️ Its `main` has **not** been updated: it still
  schedules `django_logic.recover_stranded_states` and imports it in
  `run_starter`, so that branch needs merging before anything deploys the rig
  from `main`.


emil-balashov:
## Heroku gate: PASS — 31 rows, 0 failures

Engine `3a92708` on `django-logic-test`, release **v78**, verified on the dyno before the run (`_ENGINE_PARAM_KWARGS` and `SourceStateChanged` both import, so the fleet really was on the fourth-pass engine).

| group | rows | result |
|---|---|---|
| preflight | broker, DB, cache, `STARTER_QUEUE`, dyno topology | **OK** |
| P0 | TC-01, TC-04 ⭐ crash, TC-18 ×2, TC-13, TC-12 ×2, TC-16 | **9 / 9** |
| **012 (new)** | TC-29a/b/b-ii/c (#178), TC-30a ×2/b ×2/c/d (#179) | **11 / 11** |
| P1 safety nets | TC-09 (rewritten), TC-10, TC-15, TC-11 | **5 / 5** |
| concurrency (real PG) | TC-07, TC-08 | **3 / 3** |
| durability (real beat retries) | TC-02, TC-03 | **3 / 3** |
| on-dyno suite | full pytest on a `heroku run` dyno | **85 passed** |

Every group also passed its `fleet clean` assertion — nothing left uncompleted, no instance left in an in-progress state.

**The rows that only real infrastructure could prove:**

- **TC-29c** — a NUL byte in an exception message. SQLite stores it happily; PostgreSQL rejects the write, which is precisely how the *accounting* write became the statement that failed. `record_error` survived, `errors_count` reached 1, escaped `\x00` on the row.
- **TC-30c** — a live attempt holding a genuine `select_for_update(nowait)` row lock. The watchdog deferred rather than charging it; the run completed to target with `errors_count` 0.
- **TC-07 / TC-08** — the partial-unique admission race and the phase-2 row lock, both needing two real connections.
- **TC-04** — a `SIGKILL`ed worker mid-transition, redelivered by `acks_late` + `reject_on_worker_lost`, completing exactly once.
- **TC-02 / TC-03** — retry-to-success and terminal-at-`MAX_ERRORS` driven by real beat ticks rather than an inline `retry_pending()`.

Observed live: `django_logic.watchdog_stale_attempts` firing on beat's own schedule and returning **0** while nothing is stale. That is the answer the pre-fix engine could not give — it could not see a live attempt at all, and re-charged whatever it did match on every tick.

`TC-09` was rewritten for this release: its old "tick the watchdog 5× against one `started_at`" recipe is wrong by design now that an attempt is charged at most once, so each round re-stamps.

Harness branch: `Borderless360/django-logic-test@chore/validate-0.12.0` (`docs/RESULTS.md`, the new matrix rows, `veto_states` / `nul_error` injection, and a re-entrant `simulate timeout`). Dynos parked at 0; no demo config vars left set.

CI on this commit is green across the full matrix (Python 3.11–3.14 × Django 4.2–6.0, `django-main`, both PostgreSQL stability jobs) and Bugbot found nothing this round.

Follow-ups filed rather than fixed here, none of them regressions of this branch: #184 (lock key omits the DB alias), #185 (`Action.fail_transition` TOCTOU), #186 (MTI/proxy lock + sweep keying — E001 detection landed, the key shape deliberately did not), #187 (test settings need a second DB alias to pin the routing fix).

PR #206, opened 2026-08-14
Docs: declarations are specifications — write processes and their usage out in full (#205)

### LINKED ISSUE #205: Docs: process declarations are specifications — write them explicitly; duplication is acceptable
From the gv import_app port review (Borderless360/gv#9565). A design principle the library's docs never state, learned twice now:

## The principle

A `Process` declaration is the specification of a business process. Its reader asks "how does this process behave?", not "how is this code structured?". So:

- **Write every process and every transition out in full.** Explicit `sources`, `target`, `conditions`, `side_effects`, `callbacks`, `queue` — per declaration, even when ten sibling processes repeat the same shape.
- **Duplication in declarations is acceptable, and usually preferable.** That is abnormal for "usual" code, and deliberate here: repeated declarations tell how the process works; a builder/factory/base-class that assembles transitions tells how the code works, and hides the one line the reader came to check.
- **The same rule applies at usage sites.** Code that drives a process should read `instance.process.action(...)` literally — no string-dispatch helpers, no wrapper that hides which transition runs.

## What to add to the docs

1. A short "declarations are specifications" section in the README's writing-processes guidance, with a before/after example (a transitions-builder vs. two explicitly repeated declarations).
2. A note in the nested-processes recipe: condition-disambiguated siblings that share an action name are the strongest case — each sibling's full declaration is what makes its behaviour reviewable on its own.
3. Pair it with the existing rule that side effects must not drive synchronous transitions on their own instance mid-flight: the explicit place for a data-dependent outcome ("verdict") is a named callback that runs the process after the durable row completes — show that pattern once, explicitly.

Per the release policy this is a docs-only change and consumer-driven.

emil-balashov:
Closes #205. The design principle django-logic was built on, stated in the docs for the first time — re-learned on the gv import_app port (Borderless360/gv#9565), where a transitions builder and a name-string verdict dispatcher both had to be unrolled in review.

## What it adds

- **README → Usage → "Declarations are specifications"**: write every process and every transition out in full; duplication across sibling declarations is acceptable and usually preferable — repeated declarations tell how the *process* works, a builder tells how the *code* works and hides the line a reviewer came to check. Same rule at usage sites: `instance.process.approve(...)` literally, no `getattr(process, name_variable)` dispatch. With a compact before/after.
- **README → Background → the "verdict" pattern**, placed right next to the one-in-flight gate it exists for: a background side effect cannot drive a synchronous transition on its own instance while its row is open, so a data-dependent outcome states the decided call explicitly (`partial(import_obj.process.mark_as_valid, data=rows)`) and a named callback runs it after the row completes. Includes the honest failure note (a crash between completion and callback loses the verdict, leaving the instance re-drivable, not corrupted).
- **Nested-processes recipe**: condition-disambiguated siblings sharing an action name are the strongest case — each sibling's full declaration is what makes its behaviour reviewable on its own; the repetition is the specification.
- **CLAUDE.md**: the same rule under "What to generate", where AI tools writing consumer code will actually read it.

Docs-only; suite 574 OK. Consumer-driven per the release policy.

---

> [!NOTE]
> **Low Risk**
> Documentation and version metadata only; no runtime or API changes.
> 
> **Overview**
> **Documentation-only release (0.14.1).** No library code changes; behaviour matches 0.14.0.
> 
> The docs now state a design principle that was implicit in the project: **process declarations are specifications**, not DRY code. Reviewers should see every transition spelled out (`sources`, `target`, hooks) and call sites should use literal `instance.process.action(...)`, not builders or `getattr(process, name)` dispatch. The README adds section **“Declarations are specifications”** with a before/after example, renumbers later usage sections, and documents the **verdict** pattern (`partial` on the chosen transition + a named callback after the background row completes) next to the one-in-flight gate. The nested-processes recipe adds guidance for condition-disambiguated siblings that share an action name. **CLAUDE.md** carries the same rule for AI-generated consumer code. **CHANGELOG** records 0.14.1 accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a5d5426dd596c133b513089ee02943664403248. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## [Unreleased]
 
+## [0.14.1] — 2026-08-14
+
+Documentation only. No engine change, so upgrading from 0.14.0 changes no
+behaviour.
+
+### Changed
+
+- **Docs: declarations are specifications** (#205). The README states the
+  design principle — write every process and usage out in full; duplication
+  in declarations is acceptable and usually preferable — with a
+  before/after example, the data-dependent-outcome ("verdict") callback
+  pattern next to the one-in-flight gate, a note in the nested-processes
+  recipe, and the rule in CLAUDE.md for AI tools.
+
 ## [0.14.0] — 2026-08-14
 
 ### Removed (breaking) — the second diet
@@ -67,12 +81,6 @@ growth defended the engine's own machinery).
 
 ### Changed
 
-- **Docs: declarations are specifications** (#205). The README states the
-  design principle — write every process and usage out in full; duplication
-  in declarations is acceptable and usually preferable — with a
-  before/after example, the data-dependent-outcome ("verdict") callback
-  pattern next to the one-in-flight gate, a note in the nested-processes
-  recipe, and the rule in CLAUDE.md for AI tools.
 - The testing-guide scenario catalog is five canonical scenarios; the
   other shapes stay pinned in `tests/` and the guide points there.
 - Kept, deviating from the diet plan: `retry_pending` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,12 @@ growth defended the engine's own machinery).
 
 ### Changed
 
+- **Docs: declarations are specifications** (#205). The README states the
+  design principle — write every process and usage out in full; duplication
+  in declarations is acceptable and usually preferable — with a
+  before/after example, the data-dependent-outcome ("verdict") callback
+  pattern next to the one-in-flight gate, a note in the nested-processes
+  recipe, and the rule in CLAUDE.md for AI tools.
 - The testing-guide scenario catalog is five canonical scenarios; the
   other shapes stay pinned in `tests/` and the guide points there.
 - Kept, deviating from the diet plan: `retry_pending` and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,15 @@ Use `BackgroundTransition` (durable, runs side-effects on a Celery worker,
 writes target/`failed_state`) or `BackgroundAction` (same durability, no state
 change on success) for anything slow, external, or retriable.
 
+**Declarations are specifications.** Write every process and its transitions
+out in full — explicit `sources`, `target`, `conditions`, `side_effects`,
+`callbacks` per declaration, even when sibling processes repeat one shape.
+Duplication in declarations is acceptable and usually preferable: it tells
+how the process behaves; a builder that assembles transitions tells how the
+code works and hides the line a reviewer came to check. The same rule holds
+at usage sites: drive `instance.process.action(...)` literally, never through
+name-string dispatch.
+
 ## Non-negotiable rules
 
 1. **Side-effects must be idempotent against external systems.** Background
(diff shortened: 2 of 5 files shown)


emil-balashov:
Both comments fixed — the new text was using the exact dialect it documents against, which rather proves the PR's point. `leg`, `re-drivable`, and `name-string dispatch` are gone: the verdict note now says whose row refuses the transition and what a lost callback leaves behind (the instance stays in its current state; the operator runs the action again — nothing is corrupted). The recipe note now names the case it applies to — several nested processes declaring the same action name, each selected by a condition — instead of assuming the reader knows the term, and says explicitly that the sketch above (different action names) is not that case.

emil-balashov:
Follow-through on "no remaining place": the honest census showed the ban list was not the only carrier — the language pass had deliberately left test prose and names as shipped (~170 occurrences), plus the design document's diagrams and the `Phase2 Start` log line. All of it is now plain: tests swept (one file renamed to `test_worker_state_guard.py`), the design document speaks in enqueue/execute, the worker's start log line says `Execute Start` (no consumer greps the old token — checked gv and the harness), and both rule files state the positive instruction without enumerating dead words. The one deliberate survivor is the literal settings key `PHASE2_STATE_GUARD` inside the removed-settings advice, which must match what a consumer's settings file actually says. Suite 574 OK.

emil-balashov:
Rebased onto the released `master` (0.14.0). This PR is now **+94/-4 across 5 files**, down from +830/-253 across 44.

**The third commit is dropped** (`docs+tests: retire the dialect everywhere, including its own ban list`). Three reasons:

1. **0.14.0 already did that sweep** — including the identical `Phase2 Start` → `Execute Start` change in `runner.py`. Re-applying it conflicted in ~25 files with no new content.
2. **Its rule files were thinner than the ones now on master.** Its `CLAUDE.md` and `.cursor/rules/readable-names-and-comments.mdc` would have *removed* the ASD-STE100 sentence rules, the retired-word → replacement table, and the keep-these-names-as-they-are list.
3. **It was the commit that committed `uv.lock`** (+498 lines). The repo decided that file is local tooling; it only got in because `.gitignore` on master was ignoring `.secrets/uv.lock` instead of `.secrets/` plus `uv.lock`. 0.14.0 restored both entries, so it is correctly excluded now.

Nothing was lost: `tests/test_voice.py` (the 0.14.0 dialect guard) passes on this branch, so the vocabulary this PR wanted to enforce is enforced.

What remains is exactly this PR's own content — the design principle in the README, the "verdict" callback pattern next to the one-in-flight gate, the nested-processes note, and the rule in `CLAUDE.md`. One README conflict was resolved to keep the new verdict section with master's plainer wording of the following sentence.

The changelog entry moved out of the already-released `[0.14.0]` section into a new **`[0.14.1]`**, and `pyproject.toml` is bumped — this content is not in 0.14.0.

577 tests OK · metadata drift OK · voice guard OK · `check --fail-level WARNING` clean.

emil-balashov:
The Bugbot finding is against commit `88f8ccf`, the first commit read in isolation. The second commit (`ff4fd80`, "no leg, no re-drivable, no name-string dispatch") already replaced all three terms. Verified on the branch tip — `re-drivable`, `leg` and `name-string` return zero hits across README.md, CLAUDE.md and docs/recipes/nested-processes.md, and `tests/test_voice.py` passes. No change needed.

PR #218, opened 2026-08-20
0.16.0 — the pull engine: workers claim rows from the database (#217)

### LINKED ISSUE #217: Design cut: workers pull rows from the database, so the broker mirror and its guards go away
## Summary

The background engine keeps its truth in the database row, but execution is push: every row is mirrored into a broker message, and the machinery that reconciles the two is most of the engine's bulk and most of its incident history — the dispatcher, the periodic starter, the watchdog's stamp, the stuck-detector, the beat wiring, and (0.15.0) the dispatch claim, counter, ceiling and refund. The incident list rooted in "the message and the row disagree" is long past the release policy's design-cut threshold: the silent beat-schedule merge, the interval reset (#203), lost-dispatch recovery, duplicate dispatches (#211), the unbounded queue nobody consumes (#215).

The cut: workers pull uncompleted rows from the database (`SELECT … FOR UPDATE SKIP LOCKED`), with `LISTEN/NOTIFY` as the wake-up so latency stays near push levels. The committed row is the signal. There is no message to lose, duplicate, bound, or count, and "does that queue have a consumer?" becomes a plain query.

## What it deletes, keeps, adds

Deletes: the dispatcher's broker half, the periodic starter, the dispatch claim/ceiling machinery, the beat schedule and its checks. Keeps unchanged: the declarations and the whole consumer API, the runner (savepoints, state guard, superseded rule, failure paths), sync mode, the testing package. Adds: a worker loop and a claim query — a few hundred lines against roughly a thousand deleted, and the operator's model shrinks from seven moving parts to three.

## Direction

A design record in `docs/design/` plus a working spike behind a new `BACKGROUND_EXECUTION='pull'` mode, additive and off by default, validated against the same crash table and Heroku matrix as push before any deletion happens. Celery mode stays until consumers have moved. The build-vs-adopt question (a maintained PostgreSQL job library instead of our own loop) is part of the design record.

## Found in

The 0.15.0 review cycle (#209–#215), where the fifth incident of the mirror class (#215) was fixed with more mirror-guarding machinery — correct in isolation, and exactly the growth pattern the release policy exists to stop.

emil-balashov:
## What this is now

No longer a spike — **the simplified engine, built.** Workers claim committed rows straight from the database; the broker mirror and every mechanism that reconciled it are gone. Based on the 0.15.0 branch (#216), so it merges after it and ships as 0.16.0. Closes #217.

## The design

Enqueue is unchanged: one atomic block writes `in_progress_state` and the `TransitionMessage` row. That committed row is the only signal there is — a payload-free `NOTIFY` wakes the workers, a five-second poll is the floor. A worker claims one row (`SELECT FOR UPDATE SKIP LOCKED`), runs the unchanged execute path, and asks again. The claim's filter **is** the retry rule: a failed row reappears after `RETRY_MINUTES` on its own, a running attempt is protected by its own row lock, and a crashed worker's row is claimable the instant its connection dies — faster than the starter it replaces. The worker loop also runs the safety nets once a minute, so **nothing is scheduled anywhere**: no beat, no schedule merge, no interval reset, no missing-entries check.

```
python manage.py dl_worker --queues django_logic.critical,django_logic.fast
```

## What is deleted (breaking, documented in the changelog)

- The celery dispatch, the periodic starter, `beat_schedule()`, `STARTER_QUEUE`, and the `django_logic.W002` check.
- 0.15.0's dispatch claim, counter, ceiling and refund — migration 0009 drops the two columns. A queue nobody serves is now simply a visible row: `detect_stuck_transitions` names it and its queue once it outlives the retry window, and the moment a worker starts, it takes the row.
- **Celery as a dependency.** Nothing imports it; the `[celery]` extra stays as an empty alias so old pins resolve; `BACKGROUND_EXECUTION='celery'` reports its removal with the migration steps at boot.

Net engine change against 0.15.0: **−451 lines**, and the operator's model shrinks from seven moving parts (broker, dispatcher, beat, starter, watchdog, stuck-detector, cleanup) to three (worker loop, watchdog, cleanup).

## What is unchanged

The declarations and the whole consumer API, enqueue's guards, the runner (savepoints, state guard, superseded rule, permanent failures), sync mode, and the testing package. No consumer process definition changes — migration is one setting, the worker command lines, and one queue drain.

## Validation

- SQLite suite: 597 tests OK; `makemigrations --check` clean; voice and metadata checks green.
- PostgreSQL + Redis (`tests.background` + `tests.stability`): 300 tests OK, including the pull claim under real locks: fresh row claimed and completed end to end, failed row invisible until the retry window, running attempt skipped by its lock, queue filter, exhausted-row exclusion, a full loop pass draining and running the safety nets, and the cross-process nested-restore tests re-driven through pull.
- Still to run before publish: the Heroku matrix under pull (the harness needs its Procfile moved to `dl_worker` lines), next to the push results recorded on 2026-08-20 — and the gv consumer suite against this candidate.

The design record (`docs/design/PULL_WORKERS.md`) is updated to shipped status, with the crash table mapped row by row and the build-over-adopt decision recorded.

---

> [!NOTE]
> **High Risk**
> Breaking change to the durable background engine: default execution mode, crash recovery, and operator deployment all change, and Celery is no longer a dependency. Consumers must switch workers and drain old broker queues.
> 
> **Overview**
> **Breaking 0.16.0 design cut:** background work no longer mirrors rows into a broker. Workers claim committed `TransitionMessage` rows with `SELECT FOR UPDATE SKIP LOCKED`; enqueue only `NOTIFY`s after commit. Default `BACKGROUND_EXECUTION` is `'pull'`. `'celery'` raises at boot with migration steps.
> 
> **Operator model.** Run `manage.py dl_worker --queues <names>` per queue group. The loop is the retry rule (failed rows reappear after `RETRY_MINUTES`; a dead worker’s lock dies with its connection). Safety nets (watchdog, stuck report, cleanup) run inside the loop once a minute — no beat, no `STARTER_QUEUE`, no `W002`. Attempts run in a forked child so a crash is recorded on the row and retried like a failure.
> 
> **Removed with the mirror:** Celery tasks/`beat_schedule`, dispatch claim/count/refund (migration `0009` drops `last_dispatched_at` / `dispatch_count`), and Celery as a core dependency (`[celery]` stays an empty extra). Declarations, enqueue, the runner, sync mode, and the testing package are unchanged. Pull mode still refuses SQLite and a per-process cache in production.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38b051a50d722ca98d4239abb60db0d48d7af90f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


emil-balashov:
Heroku validation of the pull engine (6337bbd) on django-logic-test, run head-to-head against the push results recorded yesterday: **33 / 33 matrix rows passed**, plus both manual rows, plus the gv consumer suite.

**The run earned its keep: it found one real design gap, now fixed in this PR.** On the first pass, the injected crash rows killed the whole `dl_worker` process — Heroku restarted it once, then the second crash inside the window put the dyno into a ten-minute restart backoff and the queue group sat parked. The old broker worker's process pool had been absorbing exactly this. The engine now runs every attempt in a forked child: a crash kills the child, the parent records `[crashed]` as an error on the row, the retry wait paces the next claim, and `MAX_ERRORS` bounds a crash loop. That makes crashing attempts counted and bounded for the first time — under push, a crash left no error anywhere. Pinned by a Postgres test that hard-kills a real forked attempt and watches it recover; the crash matrix rows then passed with two distinct attempt processes on one healthy dyno and no dyno restart at all.

The rest of the second pass, all green:

- Every row from the push matrix — durability on real one-minute retry waits (no beat exists anymore; the claim's filter is the retry rule), real-Postgres concurrency, the watchdog and stuck-finalizer rows, the 0.15.0 refusal and retention rows, fleet clean at the end.
- **Long attempt:** a 180-second attempt with a forced retry pass in the middle — the pass visits the row, the row lock holds, exactly one side-effect run.
- **No worker serves the queue:** with the critical worker at zero and a row back-dated past the window, the stuck report printed verbatim — *"…has waited 20 minutes on queue 'django_logic.critical' and no worker has picked it up — does a worker serve that queue? Start one (dl_worker --queues django_logic.critical); it takes the row at once."* — and after scaling the worker up, the row **completed 7 seconds later**, most of that dyno boot.
- **gv consumer suite: 8,096 tests OK** against this exact commit. The whole consumer diff is three glue edits (drop the beat-schedule merge, drop the retired `STARTER_QUEUE` key, delete the beat pin test) plus `dl_worker` Procfile lines at deploy time.

One migration note for the changelog reader, learned the hard way: platform config changes gate on the release phase, so the deployed old slug (which rejects `pull`) and the new code (which rejects `celery`) deadlock each other — bridge through `'sync'`, deploy, then flip to `'pull'`. The rig's results log records it.

Every gate this PR can pass has now passed. Remaining before publish: merge order with #216, the version-cut tag, and PyPI.

Issue #233, opened 2026-08-24
Squash the nine TransitionMessage migrations now


emil-balashov:
## Summary

Nine migrations, 263 lines, describe one table. The churn is visible in the files:

- `0001_initial` creates `instance_id` as a `PositiveIntegerField`; `0005` widens
  it to a `CharField`.
- `0001_initial` adds the one-uncompleted-per-instance constraint; `0006` drops it
  for the per-process one.
- `0008` adds `last_dispatched_at`, `dispatch_count` and `ended_in_failure`;
  `0009` removes the first two.

A fresh install replays six columns being added, one being widened, and two being
added and then dropped.

## Rename 0008 now

`0008_transitionmessage_dispatch_marker.py` is named after a column that no longer
exists, and its docstring says "Three nullable/defaulted columns" for a migration
that now leaves one. `ended_in_failure` is live at `models.py:69`,
`safety_nets.py:188` and `runner.py:174`.

Rename it to `0008_transitionmessage_ended_in_failure` and fix the docstring. A
migration named after a removed column is the first thing an operator reads when a
deploy stalls.

## Squash at 1.0

Replace `0001` through `0009` with one initial migration matching the current
model. I built that squash during the review and checked it: same columns,
byte-identical indexes including the per-process constraint, and
`makemigrations --check` clean.

Two things need care, and both were reproduced on Django 5.0:

- A squash named `0001_initial` that lists `0001_initial` in `replaces` removes
  itself. Give the squash a distinct name.
- An install that has not applied `0009` cannot jump to a replacing migration
  without the full `replaces` list.

State in the release note that a consumer must reach 0.16.x before 1.0.

## Related

#226 covers the 1.0 settings and restore bridges. The migration squash is not in
it.

## Found in

Simplification review at 0.16.0.

emil-balashov:
PR #236 lands the truth-telling half differently than asked: the 0008 docstring is fixed, but the file keeps its name until the 1.0 squash. gv (the one consumer) has already applied `0008_transitionmessage_dispatch_marker` on its vendored copy, and renaming an applied migration ghosts it in `django_migrations` — the renamed file replays as unapplied and fails on the existing column. The docstring now says exactly that, and the TODO.md 1.0 squash entry includes the rename (inside the squash, where the `replaces` list handles it) plus both squash hazards this issue reproduced.

emil-balashov:
[...]

PR #236, opened 2026-08-24
0.17.0 — the cleanup after the pull cut

### LINKED ISSUE #221: Finish leftover work from the 0.16.0 pull cut
## Summary

0.16.0 removed the broker. Workers claim committed `TransitionMessage` rows. The consumer API did not change.

Comments, docs, tests, and two copies of the terminal-failure path still describe the old system (periodic starter, `STARTER_QUEUE`, Celery beat, `retry_stale_transitions`). That leftover is most of the remaining size. It is not unused product features.

This issue tracks the cleanup. Do the child issues in the order listed. Stop after a child if the next one is not paying for itself. Do not add engine while cutting.

## What not to change

These stay locked:

- The cache lock is the mutex. The row is the truth.
- No busy state without a row that owns recovery.
- Workers claim with `SELECT FOR UPDATE SKIP LOCKED`.
- One uncompleted row per instance per process.
- Manual state fixes win (`[superseded]`).
- Side-effects in a savepoint. Callbacks are best-effort.
- Forked attempt processes.

## Child issues (in order)

1. #222 Remove broker-era language from docs and comments.
2. #223 Stop advertising command classes and leftover Celery pins.
3. #224 Rename pass-named test files to the contract they pin.
4. #225 Share one terminal-failure path in the background runner.
5. #226 1.0: one transition type, fewer knobs, drop migration bridges. Breaking. Later.

## Found in

Project review after 0.16.0. Library ~7,100 lines; tests ~14,800; `runner.py` 1,175; README 1,403 still showing `STARTER_QUEUE`.

### LINKED ISSUE #222: Remove broker-era language from docs and comments
## Summary

0.16.0 removed the broker, the periodic starter, beat, and `STARTER_QUEUE`. A failed row becomes claimable again after the retry wait. Nothing re-dispatches it.

Comments and docs still talk as if those parts exist. A reader who did not watch the cut cannot tell what the engine does now.

## First fix: README still shows a removed setting

`README.md` (background install example) still includes:

```python
'STARTER_QUEUE': 'django_logic.starter',
```

Copy-paste of that block puts a key the engine does not read into `DJANGO_LOGIC`. `manage.py check` reports it as `django_logic.W004`.

Delete the line. Keep the other keys that the engine still reads.

## Replace leftover names

Say what happens now. Do not say "periodic starter", `retry_stale_transitions`, Celery beat, or `acks_late` unless you are describing the old design on purpose.

Places that still use those names:

- `django_logic/background/runner.py` (several comments still say the starter re-dispatches)
- `django_logic/background/settings.py` (`retry_minutes()` docstring)
- `django_logic/testing/runner.py`, `scenario.py`, `idempotency.py`
- `docs/TESTING_GUIDE.md`, `docs/logger.md`, `docs/recipes/nested-processes.md`
- `TODO.md` ("alert when beat stops running")
- Many tests, including `tests/stability/test_periodic_tasks.py` and `tests/test_issue_fixes_engine.py`

`retry_minutes()` should say: a failed row is not claimable again until this many minutes after `last_error_dt`. The pull claim's filter is the retry rule.

`TODO.md` should say: document how to alert when `dl_worker` stops. Delete the Celery beat / `CeleryIntegration` bullet. The "Non-Celery backends (RQ, Dramatiq)" later item is also spent: there is no broker to plug.

## The old design record

`docs/design/BACKGROUND_TRANSITION_ANALYSIS.md` still describes enqueue as `apply_async` and execute as `@shared_task(acks_late=True)`. Leave the file as history. Put a notice at the top: this is the broker design 0.16.0 removed; `docs/design/PULL_WORKERS.md` is the current record.

Do not rewrite `CHANGELOG.md`. It is the historical record.

## Suggested README trim (same pass if it stays small)

The README is 1,403 lines. About 500 are the background section, which overlaps `PULL_WORKERS.md` and `TESTING_GUIDE.md`. A useful trim without a docs site: keep install, quick start, bind in `ready()`, one background example, and pointers. Do not start MkDocs here; that stays a 1.0 item.

Until 1.0, the README should also show that `Action` and `BackgroundAction` do not share a lock/gate/chain contract (sync `Action` ignores `next_transition` and skips the in-flight gate; `BackgroundAction` does not). Reviewers currently have to memorise that table.

## What not to change

No engine behaviour. No new settings. No new checks.

## Found in

Project review after 0.16.0.

### LINKED ISSUE #223: Stop advertising command classes and leftover Celery pins
## Summary

A few public names and test module docs still advertise internals or a Celery pin that 0.16.0 removed. None of this is a consumer feature.

Do this after the docs/comments pass so the remaining names match the code.

## Stop exporting command classes

`django_logic/__init__.py` puts `Conditions`, `Permissions`, `SideEffects`, and `Callbacks` in `__all__`. Consumers declare lists of functions. Nothing in this repo imports those classes as a public API.

Drop them from `__all__`. Leave the classes in `django_logic.commands` so an existing direct import still works. Do not remove the classes.

The `side_effects_class` / `callbacks_class` / `failure_callbacks_class` / `permissions_class` / `conditions_class` swap hooks on `Transition` and `Process` have one test user (`tests/test_issue_fixes_0_12.py`). Do not delete those hooks in this issue. Note them for the 1.0 trim.

## Fix leftover Celery pins in tests

`tests/test_issue_fixes_engine.py` still says in its module docstring that every django-logic Celery task sets `acks_late=True` and `reject_on_worker_lost=True`. Those tasks are gone. The remaining tests (lock release on a failed write, positional arguments, `_base_manager` restore) are real. Rewrite the docstring to name those three contracts. Do not delete the tests.

## Align the retry setting's words with the claim

If the docs pass did not already land it: `retry_minutes()` in `django_logic/background/settings.py` still says the periodic starter re-dispatches an uncompleted row. The claim's `WHERE` clause is the retry rule. Fix the docstring.

## What not to change

Do not change exception types, settings, or `Action` / `BackgroundAction` here. Those belong in the 1.0 issue.

## Found in

Project review after 0.16.0.

### LINKED ISSUE #224: Rename pass-named test files to the contract they pin
## Summary

Several test files are named after the review that found the bug, not after the contract they pin. The pins are valuable. The names hide them.

Keep every assertion. Change the names so a new reader can find the contract. Merge two classes only when they assert the same thing and you can name the surviving file.

## Files to rename

Named after a review pass:

- `tests/test_pass4_engine_pins.py` (687 lines)
- `tests/test_pass4_testing.py` (301)
- `tests/test_pass4_checks.py` (216)

Named after an issue dump:

- `tests/test_issue_fixes_0_12.py` (727)
- `tests/test_issue_fixes_engine.py`
- `tests/test_issue_fixes_testing.py`

Pick names from the contract, for example watchdog-under-lock, superseded parity, hook-signature checks, snapshot restore guards. Split a file if it pins unrelated contracts. Do not invent a test framework.

## Stale stability file

`tests/stability/test_periodic_tasks.py` still describes "Stage 2" and `retry_stale_transitions`. After the pull cut, retry is the claim's filter. Retarget this file at that filter, or fold the lock tests into the other stability files and delete it.

## Optional: split the fixture module

`tests/background/models.py` is 1,058 lines and about 47 classes (chain, nested owner, archive manager, scenario fixtures). Splitting by concern makes both the tests and the fixtures readable. Do it only if a rename already requires touching the imports. Do not add a fixture factory.

## What not to change

Do not drop a pin unless another test covers the same contract and you name that test in the pull request. These files exist because this suite has shipped assertions that passed against the defect they were meant to catch.

Do not grow `django_logic.testing` in this issue. `ProcessScenario` is the consumer product. Extra helpers stay until a consumer asks to remove them.

## Found in

Project review after 0.16.0.

### LINKED ISSUE #225: Share one terminal-failure path in the background runner
## Summary

`django_logic/background/runner.py` (1,175 lines) is the shared execute path plus restore plus two copies of terminal failure.

`_handle_failure` (worker attempt) and `_finalize_terminal_from_watchdog` (watchdog / `detect_stuck`) both:

- run the state guard and mark superseded when the instance has moved
- write `failed_state` in a savepoint and never let that write block completion
- mark the row completed even when restore or the write fails
- return enough to run `failure_callbacks` after the lock is released

That duplication is how the two paths drifted before (the safety net used to run failure callbacks against a superseded instance).

## Suggested change

One helper that, with the row already locked, does: supersede, or write `failed_state`, or complete as unrestorable. The attempt path and the safety nets both call it.

Keep `run_background_transition` as the one execute function. Do not split `runner.py` into many modules first. The duplication is the problem, not the file name.

Do this after the docs/comments pass, so the merge is reviewable against words that match the pull engine.

## Restore fallbacks (out of scope here)

`_restore` / `_find_transition` still handle pre-0.4 rows with no `field_name`, and blank `owning_process_class`. Leave them. They are 1.0 removals once no such rows remain.

## What not to change

Do not change the savepoint rule, the superseded rule, callback best-effort, or the forked attempt process. Do not add a new setting or a new safety net.

## Found in

Project review after 0.16.0. Related history: the watchdog and `detect_stuck` finalizers had to grow their own copies of `_handle_failure` after defects on the attempt path.

### LINKED ISSUE #227: Worker wake-up calls psycopg2 methods on a psycopg3 connection, so every wait costs two poll intervals
## Summary

`_wait_for_work` (`django_logic/background/pull.py:207-227`) wakes a worker when a
notification arrives. It calls `raw.poll()` and `raw.notifies.clear()`. Both are
psycopg2 methods. The project declares `psycopg[binary]>=3.1` (`pyproject.toml:60`).

Checked against psycopg 3.3.4:

- `psycopg.Connection` has no `poll`.
- `notifies` is a generator method, so it has no `clear`.
- `fileno` exists, so `select.select` still works.

The call therefore raises `AttributeError` on every wait. The bare
`except Exception` catches it and sleeps a full `timeout`.

## What it costs

The wait already spent up to `POLL_SECONDS` inside `select.select` before the
raise. The except branch then sleeps another full `POLL_SECONDS`. Every idle wait
costs two intervals instead of one.

A notification that arrives at once is worse off. `select.select` returns straight
away, `raw.poll()` raises, and the worker still sleeps the whole interval. The
notification makes pickup slower than a plain poll would.

## No test covers it

No test names `_wait_for_work`, `notify_workers` or `NOTIFY_CHANNEL`.
`tests/background/test_pull_workers.py` calls `run_worker(forever=False)`, which
returns before the wait runs.

`LISTEN` is also re-issued on every iteration. It lasts for the session.

`README.md:1208` tells consumers a lost notification costs five seconds. With the
double sleep it costs two poll intervals.

## Suggested fix

Delete the wake-up: `NOTIFY_CHANNEL`, `notify_workers`, `_wait_for_work`, and the
`transaction.on_commit(notify_workers)` in `dispatch.py`. `run_worker` then waits
with `time.sleep(POLL_SECONDS)`.

The row is the queue. The notification was only ever an optimisation, and as
shipped it is the opposite. Write it again against psycopg3's `notifies()` only if
a consumer asks for faster pickup, and only with a test that proves a notification
arrives.

## Check before landing

Run a worker against PostgreSQL and confirm the except branch is taken. This
report rests on the published psycopg 3.3.4 API, not on a live worker. A consumer
still running psycopg2 has a working wake-up today and would drop to the poll
interval.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #228: Write the savepointed failed_state write once instead of four times
## Summary

The savepointed `failed_state` write exists four times. Two copies are on the
synchronous path and two are on the worker.

- `django_logic/transition.py:326-369` — `Transition.fail_transition`
- `django_logic/transition.py:605-645` — `Action.fail_transition`
- `django_logic/background/runner.py:432-469` — the safety-net finalizer
- `django_logic/background/runner.py:838-871` — `_handle_failure`

Each is the same seven statements: read the previous state, write the new one
through `_run_in_savepoint` with `require_commit`, restore the attribute when the
savepoint is discarded, log the outcome.

I diffed them. Outside the log strings they are identical. The three-line comment
explaining why the savepoint opens on the instance's alias is written out three
times. The comment explaining `require_commit` appears in all four.

## One copy logs a different line

`runner.py:437` logs `set failed_state=...`. `runner.py:868` logs the standard
`SET_STATE` event for the same thing. A shared helper settles that.

## Why now

Three releases have hardened this shape at overlapping subsets of the four sites:
0.12.0, then the two background writers (#189), then the two sync writers (#192).
The release policy in `CLAUDE.md` already names this case. A third fix on one
defect class calls for a design cut, not a fourth fix.

## Suggested change

One helper beside `_run_in_savepoint` in `commands.py`, which both `transition.py`
and `runner.py` already import from. It reads the previous state, writes in the
savepoint, restores the attribute, logs, and returns the write error or nothing.
The four call sites become a few lines each.

Roughly 70 lines go, counting the lines the helper adds.

## Scope note

This is wider than #225, which covers the two copies inside `runner.py`. The two
`transition.py` copies are the same shape and belong in the same pass. #225 can
land first or this can absorb it.

## What not to change

Keep the savepoint rule, `require_commit`, and the alias the savepoint opens on.
Any test asserting the old `set failed_state=` wording needs updating.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #229: timeout= cannot stop a hanging attempt: bound the forked child, then drop the watchdog
## Summary

`timeout=` cannot stop a hanging attempt. It only charges one after the fact.

`abandon_timed_out_attempt` (`runner.py:187-291`) skips any row a worker still
holds (`runner.py:219-224`). Side-effects run inside that same row lock
(`runner.py:519-520`, `684-736`). So the watchdog only ever reaches an attempt
whose lock is already gone, which means a dead process or a dead connection.

An attempt that hangs while its worker lives keeps its lock and keeps running. The
keyword advertises a budget it cannot enforce.

The documents contradict each other, and one contradicts itself.

- `README.md:1215` says an attempt that hangs while keeping its connection is the
  watchdog's job.
- `docs/design/PULL_WORKERS.md:99` lists that same case as "same watchdog,
  unchanged".
- `docs/design/PULL_WORKERS.md:118` describes what the code does: `started_at` and
  `timeout_seconds` name an attempt that is abandoned **and unlocked**.

An attempt that holds its connection holds its row lock, so the watchdog skips it.
The first two statements promise something the code cannot do.

## Fix history

Five fixes across four releases:

- 0.12.0, #179 — it charged the same attempt every tick, and a fourth tick
  terminalised the row. That entry records that declaring `timeout=` made a
  transition strictly less reliable than omitting it.
- 0.12.0 — it charged and terminalised a healthy attempt milliseconds old.
- 0.12.0 — its finalizer ran destructive hooks against an operator's manual fix.
- 0.13.1 — retry classification had to read the watchdog's own signals first.
- 0.15.0, #210 — a live long attempt was reported as stranded.

## Suggested change

Bound the wait the worker parent already performs. The parent claimed the row, so
it knows `timeout_seconds`. Wait that long in `_run_attempt_in_child`, then kill
the child.

The death already has a durable path. `_record_child_death` (`pull.py:153-183`,
0.16.0) charges exactly one error in one conditional statement, so the claim
filter paces the next attempt and `MAX_ERRORS` ends it. That makes `timeout=`
mean what it says for the first time.

Then remove:

- `abandon_timed_out_attempt` (`runner.py:187-291`), including both under-lock
  re-checks, which were themselves separate fixes
- `watchdog_stale_attempts` (`safety_nets.py:74-115`)
- its step in `_run_safety_nets` (`pull.py:190-204`)
- the declared-budget branch and `RETRY_SLACK` in `retry_status`
  (`models.py:203-254`) — `models.py:203` says the slack exists to match the
  watchdog

Keep the `timeout_seconds` column and the started index. Dropping a column needs a
migration and buys nothing.

## Behaviour changes to weigh

`timeout=` starts killing attempts instead of only charging them. Anyone whose
attempts routinely run past the budget loses work they keep today.

Enforcement exists only where a child exists, so it does not apply when the
attempt runs in-process.

Durability is untouched: one error per dead attempt, recorded on the row, paced by
the claim filter, terminal at `MAX_ERRORS`.

## Check before landing

Count the consumer declarations that set `timeout=`, the same census 0.14.0 used.
If nobody sets it, the simpler option is to delete the keyword too.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #230: One tree walk can answer which background transition a row names
## Summary

Three functions answer one question: which background transition does this row
name?

- `_find_transition` (`runner.py:1049-1129`)
- `_find_background_transition_in_owner` (`runner.py:1132-1152`)
- `_background_transitions_named` (`runner.py:1155-1175`)

That is 127 lines, 59 of them docstring. The `_find_transition` docstring alone is
43 lines and re-states the architecture rule plus the history of two deploys.

All three walk the same process tree and apply the same two tests: the action name
matches, and the transition is a background one. The second walk exists only to
count distinct matches for the ambiguity check.

## One early return buys nothing

When the recorded process class is in the tree but no longer declares the
transition, `_find_background_transition_in_owner` returns nothing at
`runner.py:1150-1151`. `_find_transition` cannot tell that from a missing class,
so it logs a warning that misreports the cause and then runs the same name
fallback it would have run anyway (`runner.py:1093-1129`).

## Suggested change

One function, one walk. Iterate the tree once. For each transition whose name and
kind match, return it at once when its class path equals the recorded owner,
otherwise collect it, deduplicated by identity. After the loop: warn when an owner
was recorded and not matched, return the single match, refuse more than one,
return nothing on zero.

That is about 63 lines with the cycle guard intact, so roughly 64 lines go.

## Verified

I applied this during the review and ran the 597-test SQLite suite. It passed.
Two test files import these helpers by name and need updating:
`tests/test_pass4_engine_pins.py:651-687` and
`tests/background/test_condition_disambiguated_nested.py:28`.

## Scope note

#225 says to leave `_restore` and `_find_transition` alone because they carry
pre-0.4 row fallbacks. That is about the fallbacks, which stay. This is about the
three-way split around them.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #231: Background enqueue re-runs conditions and permissions before it takes the lock
## Summary

`BackgroundTransition.change_state` re-runs the conditions and permissions that
the resolver already evaluated.

`_iter_available_with_owner` runs the process check at `process.py:246` and the
transition check at `process.py:256-258`. `_resolve_transition_with_owner` then
requires exactly one match. Only after that does `_get_transition_method` call
`change_state` (`process.py:184`).

`background/transitions.py:137` calls `self.is_valid(...)` again with the same
user. The synchronous path does not, and the asymmetry has no stated reason.

## The re-check closes no race

It sits at line 137. The lock is taken at line 152. So it runs before the lock and
cannot close any window. It only repeats work.

Conditions are usually database reads, so every background enqueue pays a
duplicate query burst for it.

The worker does not re-evaluate conditions either. `CHANGELOG.md:1717-1724` records
that it restores the exact transition without re-evaluating. So a condition that
flips after enqueue already lets side-effects run for the whole span between
enqueue and execute.

## Suggested change

Delete `background/transitions.py:137-146`. The under-the-lock guard that matters,
`_ensure_db_state_in_sources` at line 172, stays.

## Verified

I deleted it and ran the 597-test SQLite suite. Exactly one test fails:
`tests/test_lock_logging.py:103-124`. Its own comment at lines 104-107 says the
resolver already filters, and it reaches the branch by calling `change_state`
directly at line 118 rather than through the process. Delete that test with the
branch.

## Behaviour change

If a condition flips between resolution and enqueue, the row is now created and
the worker runs the side-effects. That is a real change, not a no-op. It is the
behaviour the synchronous path already has.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #232: Hold every setting in one module
## Summary

`django_logic/conf.py` and `django_logic/background/settings.py` are one settings
module split in two. The split is already crossed in both directions.

- `background/settings.py:17` imports the private `_conf` out of
  `django_logic.conf`, with the comment "The core reader, shared rather than
  re-implemented".
- `background/settings.py:144-145` calls `conf.validate_core_settings()` inside
  `validate_on_ready()`, so the two boot gates are one gate with two names.
- All ten keys come from the same settings dictionary.
- `STRICT_KWARGS_SERIALIZATION` lives in `background/settings.py` while
  `STRICT_HOOK_SIGNATURES` lives in `conf.py`, for no stated reason.

Both `AppConfig.ready()` hooks also call `install_legacy_exception_base()` and
import the checks module. `background/apps.py:16-17` justifies the copy as
covering an install where only the background app's `ready()` runs, which
`README.md:137-143` does not document.

## Suggested change

Move every reader into `django_logic/conf.py`: `background_execution`,
`default_queue`, `max_errors`, `retry_minutes`, `cleanup_days`,
`strict_kwargs_serialization` and the two mode constants. One module then owns
every key, one reader, one number validator, one bool validator, and one place
where each default is written down.

Roughly 25 lines go, and the duplicated validators go with them.

## Careful with the boot gate

`tests/test_core_settings_validation.py:66` pins that the core app runs the core
gate. If the six background keys move into that gate, a synchronous-only install
starts failing at boot on keys it never set.

So the two pull-mode gates must stay in the background app:
`_reject_sqlite_in_pull_mode` and `_check_lock_cache_in_pull_mode`.

## Import path

`django_logic.background.settings` is importable. It appears in no `__all__` and
nowhere in `README.md`. Five test modules import it directly and need updating.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #234: Comments re-tell the changelog: process.py is 37% prose, models.py 50%
## Summary

Comments in three modules re-tell the changelog instead of stating the rule.
`CLAUDE.md` already sets the standard: a comment gives a non-obvious why in one or
two sentences, and `CHANGELOG.md` owns the history.

Measured with a tokenize pass:

| File | Prose lines | Code lines | Prose share |
|---|---|---|---|
| `django_logic/process.py` | 250 | 354 | 37% |
| `django_logic/background/models.py` | 209 | 166 | 50% |

This is separate from #222, which covers broker-era wording. This is about volume
and about narrating past defects.

## The heaviest blocks

`_validate_unique_background_action_names` (`process.py:373-445`) is 73 lines: 36
docstring, 6 comment, about 12 lines of logic, and a 16-line exception message.
Lines 386-408 enumerate what the validator does not reject, with two worked
examples. `CHANGELOG.md:1740-1754` already carries that explanation nearly
sentence for sentence.

`_validate_action_names_not_shadowed` (`process.py:325-370`) spends lines 338-345
re-litigating a later correction, which is `CHANGELOG.md:834-836`.

`bind_model_process` (`process.py:554-642`) is 89 lines with 27 of comment,
including a 16-line block describing two earlier wrong implementations.

`process.py:533-538` is a tombstone comment for removed code.

## Suggested change

Keep every invariant. Cut the narrative around it. Roughly 140 lines go across the
two files.

## Care needed

This is judgement, not mechanics. These blocks are institutional memory. Whoever
cuts them keeps the sentence stating the rule and drops only the story. Two of the
validators were narrowed after they shipped, so the "why we do not check this"
sentence is worth one line even when the story around it goes.

`tests/background/test_validation.py:178-179` asserts loosely on the 16-line
exception message. Check it before shortening the message.

`django_logic/background/exceptions.py` is 50 lines and almost all docstring, but
leave it alone. Those docstrings carry operator guidance, including the warning
about swallowing `AlreadyInProgress` while an attempt has already read its inputs.
That is a hazard worth the space.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #235: Redistribute dispatch.py and clear the dead names around it
## Summary

`django_logic/background/dispatch.py` is 110 lines holding three unrelated things.
None of them dispatch anything now that the broker is gone.

- `dispatch_transition` (`53-67`) is a seven-line if-else with one caller,
  `transitions.py:204-205`.
- `retry_pending` (`70-78`) is a two-line delegate to `safety_nets.run_pending`.
- `in_flight` (`81-110`) is 30 lines, 19 of them docstring, delegating to
  `TransitionMessage.retry_status`. It has nothing to do with dispatch.
- `_current_mode` (`47-50`) is read twice, and both readers compare it to one
  string, so the mode concept has collapsed to a boolean.

`transitions.py:3-16` already documents the mode boundary at the caller, so the
module docstring largely repeats it.

Nothing here is dead. Every member has a live caller. The module name no longer
describes any of them.

## Suggested change

Delete the module and place each part where it belongs.

- `sync_execution` and a boolean mode reader go beside `background_execution()`.
- The two-branch handoff inlines into `BackgroundTransition.change_state`, where
  the row is already in hand.
- `in_flight` goes beside `retry_status` in `models.py`, with its docstring
  trimmed.
- The public map points `retry_pending` straight at `safety_nets.run_pending`.

The documented import paths keep working through the lazy public map. One test and
one line of `docs/TESTING_GUIDE.md` import the module directly.

## Dead names to clear in the same pass

- `process._RESERVED_KWARGS` is read by no code. Two comments point at it as though
  it were live, at `transition.py:51` and `process.py:127`.
- `run_pending(queues=...)` is never passed by any caller. All seven call sites
  use `run_pending()`.
- `run_once`'s `isolate=False` default is never used by the library. `pull.py:241`
  passes `isolate=True`; only one test takes the default.
- About eleven unused imports across the library, two left over from removed
  features, one a shadowing re-import.
- `observability.task_label` names a task that no longer exists. No library code
  calls it — only its own test and a docstring example.
- `State.unlock()` has a token-less branch that force-deletes another holder's
  lock. No caller, no test, not documented.

## Found in

Simplification review at 0.16.0.

### LINKED ISSUE #237: Production is always pull: BACKGROUND_EXECUTION is not a deploy switch
## Summary

A `BackgroundTransition` has one meaning: write the row, a worker claims it.

`DJANGO_LOGIC['BACKGROUND_EXECUTION']` still lets a deploy pick `'sync'`. That runs side-effects in the web process. gv exposes the same knob as `DJANGO_LOGIC_NEW_EXECUTION`.

`DEBUG=False` is the wrong gate. Django tests already run with `DEBUG=False`.

## Suggested change

Production is always pull. Sync cannot turn on from a setting or an env var.

The inline path is test-only and must opt in through `django_logic.testing`:

- `sync_execution()` for one block (already used by `ProcessScenario`)
- `django_logic.testing.enable_sync()` in test settings so the suite does not wrap every call

Boot refuses `BACKGROUND_EXECUTION='sync'` unless that opt-in ran.

Stop advertising sync as a workaround. The SQLite-in-pull error currently says "or set BACKGROUND_EXECUTION='sync'". Tell the operator to use PostgreSQL.

Drop `BACKGROUND_EXECUTION` from the README install example. The testing guide is the only place that shows the opt-in.

Do this now. Do not wait for a 1.0 release.

## Consumer

gv still defaults to `'sync'` and overwrites production to `'celery'`. That is Borderless360/gv#9659.

### LINKED ISSUE #238: README should tell the reader what to do, not what the library used to do
## Summary

The README is 1,400 lines. A new reader cannot tell how to install the library, because the install section still narrates releases.

The extras block is the type:

```
pip install django-logic[redis] — installs django-redis, for deployments
whose settings name django_redis.cache.RedisCache. It stopped being a
core dependency in 0.11.0, because the engine has never imported it.
[celery] remains an empty alias, so existing pip install
django-logic[celery,redis] pins keep resolving — 0.16.0 removed the
broker, so nothing imports celery
```

The operator needs: `pip install django-logic`. Optionally `[redis]` if settings name `django_redis`. `[celery]` is not a feature. Version history belongs in CHANGELOG.md.

#222 asked for a trim and to drop `STARTER_QUEUE`. The 0.17 PR drops that key. The extras paragraph, the 0.2–0.4 upgrade note, and the rest of the history stay.

## What a first-time reader meets

- Features with emoji. "Sync execution mode" listed as a product feature.
- Requirements that treat pull and sync as two equal runtimes. Sync is tests.
- Installation that points at 0.4 and 0.2.
- Quick Start, then Usage repeats it, then a Complete Example repeats it again.
- Bind-in-`ready()` explained in five places.
- "Declarations are specifications" is assistant guidance, not operator docs.
- Django-Logic vs Django FSM in the table of contents.
- Background Transitions (~500 lines) retells `docs/design/PULL_WORKERS.md` and `docs/TESTING_GUIDE.md`. The install dict lists every knob, including flags defaulting to False and `LEGACY_EXCEPTION_BASE`. The text says an empty dict is a valid production start.
- `BACKGROUND_EXECUTION='pull'` in the install example. Pull is the default. Putting it next to a "set sync in tests" comment makes a deploy switch.
- Support still points at the GitHub wiki.

## Suggested change

Keep: install, bind in `ready()`, one process, one background example, `dl_worker`, PostgreSQL, a cross-process cache, a pointer at the testing guide and `PULL_WORKERS.md`.

Drop from the README: release numbers, empty extras as a story, sync as a product mode, the fork-migration setting, the Django FSM comparison, the assistant-style declaration rules, and the repeated bind lecture. One bind section is enough.

The `[celery]` extra can stay in `pyproject.toml` so old pins resolve. Do not mention it in the README.

Related: #222 (broker-era words), #237 (sync is tests only).

### LINKED ISSUE #239: Timeout kill can raise ProcessLookupError and stop the worker
## Summary

After `WNOHANG` says the attempt is still running, the process can exit before `os.kill`. `os.kill` then raises `ProcessLookupError`. That exception leaves the worker loop.

A successful attempt that finished on time can already have completed the row. The worker still dies.

## Suggested change

Catch `ProcessLookupError`. Reap with `waitpid` if needed. Treat a vanished process as a normal exit, not a timeout, unless `waitpid` shows a kill.

`django_logic/background/pull.py` in `_wait_for_attempt_process`.

### LINKED ISSUE #240: Leftover names after the 0.17 cut
## Summary

The 0.17 rename pass left names that still hide the work.

- `_record_error_if_uncompleted` names the `is_completed=False` guard. Callers already pass `[timeout]` / `[crashed]`. Name it `_record_attempt_error`.
- `in_flight` is a metaphor. The library word is uncompleted. gv reads `TransitionMessage.in_flight_for`; keep a public alias if you rename.
- `TestPeriodicStarterContract` and several tests still say the periodic starter re-dispatches. The claim filter is the retry rule.

## Suggested change

Rename the private helper. Replace "periodic starter" in tests with the claim. Review `in_flight` before renaming the public probe.

### LINKED ISSUE #241: CLAUDE.md, pyproject, and README extras still describe celery and deleted files
## Summary

Assistants and installers still see the broker-era product.

- `CLAUDE.md` lists `dispatch.py` and `settings.py`. Both are gone. It still says to read `BACKGROUND_TRANSITION_ANALYSIS.md` first. That file is the broker design. Point at `PULL_WORKERS.md`.
- `pyproject.toml` keywords include `fsm` and `celery`. The description says "state machines".
- README extras narrate 0.11.0 and 0.16.0. `[celery]` can stay in `pyproject.toml` for old pins. The README should not mention it.
- README Support still points at the GitHub wiki.
- `docs/TESTING_GUIDE.md` still says "timeout watchdogs".

Related: #238 (README trim), #222 (broker-era words).

### LINKED ISSUE #245: Lock key must include the database alias and the concrete model
## Summary

Today one row on two databases shares a lock key. A per-tenant-database consumer gets cross-tenant "State is locked" refusals. An MTI parent and child sharing one state column lock under different keys.

## Suggested change

Change the key once. Add the database alias. Key the model identity on the concrete table. Record the alias on `TransitionMessage` and route `_restore` by it.

The key is shared-cache identity across rolling deploys, so the release note must say: drain, or accept one TTL window.

Do this now. Do not wait for a 1.0 release.

### LINKED ISSUE #246: Operator commands: list uncompleted rows and re-send one now
## Summary

An operator during an incident cannot ask the library which rows are uncompleted or stuck, and cannot send one row to a worker now.

## Suggested change

One management command that:

- lists uncompleted rows, and names the ones at `MAX_ERRORS` or past the retry wait with no attempt started
- sends one `TransitionMessage` to a worker immediately, ignoring the `RETRY_MINUTES` wait

Do not add a second scheduler. The worker still claims the row.

### LINKED ISSUE #247: Document the Postgres connection budget and how to alert when dl_worker stops
## Summary

Each running attempt holds one database connection, or two when the app opens a second one. Nothing in the docs tells an operator how to size `concurrency × workers` against the plan cap or pgbouncer.

The safety nets run inside the worker loop. A dead `dl_worker` also means no stuck finalizer and no cleanup. Nothing says how to alert on that.

## Suggested change

One short ops page, or a README production subsection:

- connection budget
- alert when `dl_worker` stops (process supervisor, Sentry cron, health check)

No new engine.

### LINKED ISSUE #248: Handled outcomes should not log at ERROR
## Summary

`detect_stuck` completing a row, the timeout kill, and the unrestorable path all log at ERROR. Those are outcomes the engine already handled. They fill Sentry.

## Suggested change

Log them at WARNING or INFO. Keep ERROR for cases the operator must act on (no worker serves this queue, a notification wait failed in a loop, a settings boot failure).

### LINKED ISSUE #250: TransitionNotAllowed should name the current state and the available actions
## Summary

`TransitionNotAllowed` does not say what state the instance is in, or which actions are available. The operator or the API layer has to reconstruct that.

## Suggested change

Put current state and available action names on the exception. Keep the type. Do not change HTTP mapping (`TransitionTemporarilyUnavailable` vs `TransitionNotAllowed`).

### LINKED ISSUE #251: Publish to PyPI automatically on tag push
## Summary

Releases are cut by hand. A git tag should publish the package.

## Suggested change

A CI job on tag push: build, twine upload, using a trusted publisher or a stored token. Do not publish from `master` without a tag.

### LINKED ISSUE #257: dl_worker runs one attempt at a time, so capacity is process count and memory is reserved for the worst case
## Summary

`manage.py dl_worker` runs one attempt at a time: `run_once` claims one row, forks one child, waits for it, asks again. A deployment's only throughput lever is more worker processes.

On Heroku that lever is expensive in exactly the wrong dimension. Each dyno is a fixed memory slice, and each worker carries the full Django base image (~250 MB) plus the peak of its heaviest forked child. Memory demand is spiky — most attempts need little, a few need a lot — so fixed slices must all be sized for the peak, while one shared pool could absorb it.

## Production evidence (Borderless360/gv, 2026-08-24, first day on 0.16 pull mode)

- Six queues, three worker processes on 512 MB dynos. Two of the three hit the platform's memory-quota error within hours — an import child peaked at 804 MB, and two Shopify `fetch_orders` workers swapped at the same time.
- Swapping was the throughput failure, not just a warning: an hourly fan-out of ~119 `fetch_orders` rows drained at ~2.6 rows/minute across two swapping workers — the oldest row waited over 10 minutes.
- The fix was to double every dyno's memory. That doubles the fleet cost to buy headroom each worker only needs occasionally, and the parallelism is still fixed at one attempt per process.

## Suggested change

A `--concurrency=N` option on `dl_worker`: one parent, up to N forked attempts in flight, each child claiming and executing its own row exactly as today (`SKIP LOCKED` already makes concurrent claims safe). The safety nets keep running once per loop in the parent. Default stays 1, so nothing changes for existing deployments.

With that, an operator can run one large worker with a shared memory pool — N cheap slots that share headroom — instead of N processes that each reserve the worst case. Queue groups stay possible for SLA isolation; they just stop being the only way to get parallelism.

## Relation to #237

Independent, but they compose: #237 makes pull the only production mode; this makes a pull deployment's capacity a per-process setting rather than a process-count setting.

emil-balashov:
## Summary

0.16.0 removed the broker; this release removes what the cut left behind, then fixes the two things the consumer's first production week on pull mode actually broke on. The words now match the pull engine, one copy exists of what had four, and three mechanisms that could not do their jobs were replaced by ones that can.

Every issue below was verified against the code and against gv (the only consumer) before landing.

## What changed

**The cleanup after the cut**

- **The wake-up works on psycopg 3** — `_wait_for_work` called psycopg 2 methods on the psycopg 3 connection the project pins, so every wait slept two poll intervals. Fixed for both drivers, `LISTEN` once per session, validated against PostgreSQL 17. Pickup measured at 0.9 s against the 5 s poll floor, and that number is now in the design record.
- **`timeout=` means what it says** — the worker kills an attempt process that runs past its budget; the watchdog (which could only reach already-unlocked rows, and produced five fixes across four releases) is gone. Zero `timeout=` declarations exist in gv.
- **Background enqueue stops re-running conditions** before it takes the lock; the under-lock source guard is unchanged.
- **One copy of what had four** — the savepointed `failed_state` write, the terminal completion, and the row-to-transition tree walk.
- **One module per concern** — `conf.py` owns every settings key; `background/settings.py` and `background/dispatch.py` are gone.
- **`DEFER_UNLOCK_UNTIL_COMMIT` is deleted** (651 net lines): nothing set it.
- **The README is 363 lines, down from 1,387** — install, bind, one example each, the commands, the two safety nets, one pointer per topic.

**What the consumer's production week added**

- **A worker runs several attempts at a time** (#257). `dl_worker --concurrency=N`, default 1. Capacity used to be process count alone, which on a platform selling fixed memory slices means every one-slot worker is sized for its heaviest attempt. gv's first day on 0.16.0 lost two of three workers to the memory quota and drained an hourly fan-out of 119 rows at 2.6 rows a minute. The claim now also skips the rows the worker already runs — a worker claims faster than its attempt processes take their row locks, so without that the head row filled every slot.
- **An operator command** (#246). `dl_transitions` lists every uncompleted row and says why each is not moving; `--send <pk>` clears one row's retry wait and wakes the workers. During the gv incident nobody could ask the library either question.
- **Sync is a test runtime, not a deploy switch** (#237). Boot refuses `BACKGROUND_EXECUTION='sync'` unless `conf.enable_sync()` ran first, which a test settings module calls. This is the library half of Borderless360/gv#9659.
- **The state lock names the row, not the class** (#245). The key was built from the model name, so a proxy model and the model it proxies took two locks on one row and one state column. gv has four processes bound to one physical export table through proxy models. The key now names the table that holds the column.

**Two defects this review round found in its own fixes**

- A killed attempt reaped elsewhere is a timeout, not a success — and an attempt that ended on its own just before the kill is no longer charged one. `os.kill` succeeds on a process waiting to be reaped, so the kill alone cannot decide the outcome; the reaped status does.
- A refusal stays a refusal. Building `available_actions` for the enriched `TransitionNotAllowed` message runs every transition's conditions and permissions, so a consumer condition that raised replaced the refusal with its own exception — a 500 where the caller expected an answer.

## Upgrade steps

1. Test settings that set `BACKGROUND_EXECUTION='sync'` must call `django_logic.conf.enable_sync()` beside it.
2. The state lock key changed, and it is shared-cache identity — but this needs **no drain**. The cache lock covers only the enqueue critical section; the guards that outlast it are the partial unique index on uncompleted rows and the worker'''s `select_for_update(nowait=True)`, neither of which reads the cache. A rolling deploy'''s split window weakens only the synchronous fast mutex, for the minutes the release takes.

## Validation

- SQLite suite: **590 tests OK**
- PostgreSQL 17 (`tests.background` + `tests.stability`): **309 tests OK**, including the live NOTIFY wake-up, a real timeout kill, and a three-slot concurrency proof whose side-effect only completes when all three attempts stand in it at once.
- The Heroku validation matrix is the release gate and runs before the PyPI cut.

## The issue tracker after this

Closed with this PR, or answered and closed on their own: #221 #222 #223 #224 #225 #227 #228 #229 #230 #231 #232 #234 #235 #237 #238 #239 #240 #241 #244 #245 #246 #247 #248 #249 #250 #251 #255 #256 #257.

Left open on purpose, each with its gate written on the issue: **#242** and **#233** (the one-app fold and the migration squash — gated on a migration rehearsal against a production-sized copy of the consumer's table, and deliberately not in the release the consumer takes straight after an incident), and **#243** (one transition type — a 105-declaration consumer port, which should ride with the `STRICT_*` default flip from #244 once the consumer's coupled-core ports finish).

Fixes #221
Fixes #222
Fixes #223
Fixes #224
Fixes #225
Fixes #227
Fixes #228
Fixes #229
Fixes #230
Fixes #231
Fixes #232
Fixes #234
Fixes #235
Fixes #237
Fixes #238
Fixes #239
Fixes #240
Fixes #241
Fixes #245
Fixes #246
Fixes #247
Fixes #248
Fixes #250
Fixes #251
Fixes #257

---

> [!NOTE]
> **High Risk**
> Touches the worker loop, timeout enforcement, state-lock identity, and boot-time execution mode — all production-critical paths with rolling-deploy and test-settings upgrade notes.
> 
> **Overview**
> **0.17.0** is the cleanup after the 0.16 pull cut: leftover broker-era code is gone, and the worker actually enforces what it used to only document.
> 
> **Worker behaviour.** `timeout=` now **kills** the forked attempt process (the old watchdog is removed). `dl_worker --concurrency=N` runs several attempts in one process (default still 1); claims skip rows the worker already holds. LISTEN/NOTIFY wait is fixed for psycopg 3. `dl_transitions` lists stuck rows and `--send` clears one retry wait.
> 
> **Guards and API.** Sync mode is test-only: boot refuses `BACKGROUND_EXECUTION='sync'` unless `enable_sync()` ran first. The state lock keys the **table + column + pk**, so proxy/MTI models no longer take two locks on one row. `DEFER_UNLOCK_UNTIL_COMMIT` is deleted. Settings live in `conf.py`; `dispatch.py` / `background/settings.py` are gone. Command classes drop from the top-level `__all__`.
> 
> **Docs/release.** README is cut to install + examples + pointers. A `v*` tag publishes to PyPI via trusted publishing after tests and a version/tag check.
> 
> **Upgrade:** test settings that set sync must call `enable_sync()`. Lock-key change needs no drain; rolling deploys only weaken the cache mutex for a few minutes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a63d4b581be9b17f6fc5c0924c8d17da8230f67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


emil-balashov:
**The Heroku validation gate has now passed on this branch** (engine @ afc75779, harness `django-logic-test` branch `candidate/0.17.0`, run 2026-08-24, recorded in the harness `docs/RESULTS.md`):

- **Full matrix: every row green.** 28/29 on the first pass; the single failure was the harness's own TC-08 fixture, which staged its row-lock hold with a long sleep on `payments.charge` — the 0.17.0 parent now kills that at its 15s budget, which is the new semantics working as designed. Refit (race inside the budget; hold the lock on a transition without `timeout=`) and the rerun passed 3/3.
- **The timeout kill, end to end on real dynos:** a stalled attempt killed at its budget → one `[timeout]` error → paced retry → charged; an always-stalling attempt killed five times → terminal `failed_state` via the stuck finalizer; a slow attempt inside its budget reaped clean with zero errors.
- **The wake-up, measured on psycopg 3.3.4:** enqueue → `fulfilled` in **0.9s** against a 5s poll floor. On 0.16.0 this stack silently degraded to polling — the fix is visible as a ~5s drop in pickup latency.
- **The long-attempt protection after the budget-branch removal:** two forced retry passes during a 180s attempt; `physical_runs` = exactly 1 — the running attempt's own row lock is the protection.
- Local harness suite: 76 passed.

Ready to merge and cut to PyPI.

django_logic/background/pull.py
```diff
@@ -138,19 +148,58 @@ def _run_attempt_in_child(pk: int) -> None:
...
+            return os.waitstatus_to_exitcode(raw_status), False
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            os.kill(child, signal.SIGKILL)
+            os.waitpid(child, 0)
+            return -signal.SIGKILL, True
+        time.sleep(min(1.0, max(0.01, remaining)))
+
+
+def _record_child_death(pk: int, message: str) -> None:
```
emil-balashov: wtf is the name?

django_logic/background/apps.py
```diff
@@ -1,4 +1,94 @@
...
+        _check_lock_cache_in_pull_mode()
+
+
+def _reject_sqlite_in_pull_mode() -> None:
+    """SQLite doesn't support ``select_for_update(nowait=True)`` nor
+    partial unique indexes, so the worker concurrency guard silently
+    degrades to "serialize everything" — which masks real bugs in dev
+    and fails in prod.
+
+    Only the alias that actually stores ``TransitionMessage`` is checked:
```
emil-balashov: do we even need support of sqlite? would that make things easier if we remove it?

django_logic/background/models.py
```diff
@@ -420,3 +393,27 @@ def record_failure_side_effect_error(
...
+    probe and 400 on plain ``TransitionNotAllowed`` stays consistent.
+
+    Answers ``False`` without a query when ``django_logic.background``
+    is not installed.
+    """
+    from django.apps import apps
+
+    if not apps.is_installed('django_logic.background'):
+        return False
+    return (
```
emil-balashov: django logic background is default functionality we don't need to "isntal" it

django_logic/background/pull.py
```diff
@@ -138,19 +148,58 @@ def _run_attempt_in_child(pk: int) -> None:
...
+
+    With no declared budget the wait is plain and unbounded. With one,
+    the parent polls the child and kills it (``SIGKILL`` — the attempt
+    may hang inside code that ignores gentler signals) when the budget
+    passes. The kill releases the child's row lock with its connection.
+    """
+    if timeout_seconds is None:
+        _, raw_status = os.waitpid(child, 0)
+        return os.waitstatus_to_exitcode(raw_status), False
+    deadline = time.monotonic() + timeout_seconds
```
emil-balashov: what child are you waiting for? use better names

emil-balashov: Renamed in 10df501, and the whole child/parent vocabulary went with it (it dated from 0.16.0). The forked process is the **attempt process** everywhere now: `_run_attempt_process` forks it, `_wait_for_attempt_process(attempt_pid, timeout_seconds)` waits for it and kills it past its budget, and the prose in the library, the docs, and the tests says worker / attempt process instead of parent / child.

django_logic/background/pull.py
emil-balashov: Renamed in 10df501 to `_record_error_if_uncompleted(pk, message)` — the name now states the rule (record one error on the row unless the row completed first) instead of a metaphor. The callers say what happened in the message they pass: `[crashed] the attempt process died (exit N)` or `[timeout] the attempt ran past timeout=Ns and was stopped`.

django_logic/background/models.py
emil-balashov: Agreed — removed in 10df501. `in_flight` reads the row unconditionally now; the `apps.is_installed` branch and the test that pinned it are gone. Taking the point further: TODO.md's 1.0 section now has a **One app** entry (701ff62) — `INSTALLED_APPS` should list one entry, not `django_logic` plus `django_logic.background`. It is breaking (the background app owns the `django_logic_background` app label and its migrations), so it rides with the 1.0 migration squash.

django_logic/background/apps.py
emil-balashov: Checked it: removing SQLite would buy almost nothing and cost the zero-services test story. This boot check plus one docstring sentence is the entire SQLite-specific surface of the engine — nothing else branches on the backend. What SQLite carries is sync mode: the library's own 583-test suite and every consumer's process tests run on sqlite with no services (that is the TESTING_GUIDE contract), and gv's CI benefits from the same. Pull mode already refuses SQLite here, which is exactly what keeps the rest of the engine backend-agnostic. So: keep sync-on-SQLite, keep this check. If you want it anyway, it is a one-file deletion plus a hard PostgreSQL requirement in the docs — but I recommend against it.

emil-balashov:
Review round applied and re-validated on the rig (head 701ff62):

- **One name for the forked process: the attempt process.** `_run_attempt_process`, `_wait_for_attempt_process(attempt_pid, timeout_seconds)`, and `_record_error_if_uncompleted` (it records one error on the row unless the row completed first — the name states the rule). The parent/child vocabulary is gone from the library, the docs, the tests, and the harness.
- **`in_flight` reads the row, always** — the not-installed branch and its pin are deleted; the background engine is default functionality. The full fold to one `INSTALLED_APPS` entry is now a 1.0 TODO entry (it owns the app label and migrations, so it rides with the migration squash).
- **`_finalize_stuck_row`** replaces `_finalize_terminal_from_safety_net`, and its always-'detect_stuck' source parameter is gone.
- **SQLite stays**, answered on the thread: this boot check is the engine's entire SQLite-specific surface, and what SQLite carries is the zero-services sync-mode test story (the library's own suite and every consumer's process tests).

Suites on the renamed head: SQLite and PostgreSQL green; harness re-run on Heroku — smoke 3/3, concurrency 3/3, P1 6/6 including both timeout-kill rows.

emil-balashov:
One more simplification pass over the whole library (head a0a2dfa): three parallel module reviews plus a lint sweep produced 20 candidates; 13 survived verification and are applied — **125 net lines out, zero behaviour change**, both suites green. Library is now 6,912 lines.

What went: the last four two-copy shapes (the stuck finalizer's kwargs decode, `_restore`'s recorded-class load, `_handle_failure`'s double outcome, the duplicated no-transition message), dead generality (`db_safe_text(limit=)` that nothing passed, a shadowed exception-class branch, a local re-import, the bespoke `'celery'` removal message — that migration is complete, and an unknown mode still fails loudly naming the valid ones), and the remaining history-narrating prose in checks/transition/commands/state and the testing package.

Three findings I deliberately did **not** apply, for your judgment:
1. `_run_success_hooks` runs callbacks and next_transition in two near-identical blocks (~8 lines). A loop would fold them but blands the per-hook wording.
2. The six testing assertions (`assert_available`, `assert_side_effects_ran`, …) share one body shape (~28 lines). A private helper would fold them but each bespoke failure message is operator-facing quality.
3. `snapshot()` writes a best-effort `process` block (`class`, `available_actions`, `is_locked`) that `from_snapshot` never reads and no doc mentions (~11 lines) — but it has diagnostic value in a production capture, so it is consumer-visible output and your call.

Say the word on any of the three and I cut them too.

django_logic/background/pull.py
```diff
@@ -102,61 +101,113 @@ def run_once(queues: list[str], *, isolate: bool = False) -> bool:
...
+        _, raw_status = os.waitpid(attempt_pid, 0)
+        return os.waitstatus_to_exitcode(raw_status), False
+    deadline = time.monotonic() + timeout_seconds
+    while True:
+        pid, raw_status = os.waitpid(attempt_pid, os.WNOHANG)
+        if pid != 0:
+            return os.waitstatus_to_exitcode(raw_status), False
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            os.kill(attempt_pid, signal.SIGKILL)
```
emil-balashov: **[bug]** After `WNOHANG` says the attempt is still running, the process can exit before `os.kill`. `os.kill` then raises `ProcessLookupError` (ESRCH). That exception leaves `_run_attempt_process` and kills the worker loop. The race sits exactly at the timeout boundary, including an attempt that finished on time. A successful attempt can already have completed the row; the worker still dies.

**Suggestion:** Catch `ProcessLookupError` (and `ESRCH`). Reap with `waitpid` if needed. Treat a vanished process as a normal exit, not a timeout, unless `waitpid` shows a kill.

django_logic/background/pull.py
```diff
@@ -102,61 +101,113 @@ def run_once(queues: list[str], *, isolate: bool = False) -> bool:
...
+            os.waitpid(attempt_pid, 0)
+            return -signal.SIGKILL, True
+        time.sleep(min(1.0, max(0.01, remaining)))
+
+
+def _record_error_if_uncompleted(pk: int, message: str) -> None:
```
emil-balashov: **[suggestion]** The human asked what `_record_child_death` was. `_record_error_if_uncompleted` is longer and still opaque. It does not say timeout or crash. It names the `is_completed=False` guard. The callers already pass `[timeout]` / `[crashed]` in the message.

**Suggestion:** Name the work: `_record_attempt_error(pk, message)`. Drop "if uncompleted" from the name; the UPDATE already carries that guard.

django_logic/background/models.py
```diff
@@ -420,3 +391,20 @@ def record_failure_side_effect_error(
             f'{existing}; {note}' if existing else note
         )
         self.save(update_fields=['failure_side_effect_error', 'modified'])
+
+
+def in_flight(instance, process_name: str = 'process') -> bool:
```
emil-balashov, line 399: **[suggestion]** The human said the background engine is default and must not be installed separately. This function stopped special-casing `apps.is_installed`. That is not the ask. `INSTALLED_APPS` still needs `django_logic` and `django_logic.background`. The README still says to add the second app. Folding the app label is breaking; saying the comment is done is a dodge.

**Suggestion:** Either fold the app in this release, or leave `in_flight` alone and say clearly that two apps remain required until 1.0. Do not present the `is_installed` delete as the fix.

README.md
```diff
@@ -1094,7 +1107,6 @@ A queue is a column on the row, and each worker process names the queues it serv
 
 Retries need no scheduler: a row whose attempt failed becomes claimable again after `RETRY_MINUTES` — the claim's own filter is the retry rule. Three safety nets run inside every worker loop, once a minute, so nothing else has to be configured:
```
emil-balashov: **[suggestion]** This PR deleted the watchdog bullet. The sentence above still says "Three safety nets" and then lists two (`detect_stuck_transitions`, `cleanup_completed_transitions`).

**Suggestion:** Write "Two safety nets" or drop the count.

django_logic/background/runner.py
```diff
@@ -416,62 +290,64 @@ def _finalize_terminal_from_watchdog(
...
+    kwargs: dict,
+    exception: BaseException,
+    *,
+    prefix: str,
+    consequence: str,
+    measure_duration: bool,
+):
+    """The one terminal-failure completion: write ``failed_state`` (when
+    declared), then mark the row completed. The worker attempt path and
+    the stuck finalizer both end here — the two used to carry their
```
emil-balashov: **[suggestion]** #234 asked comments to state the rule and leave the story in the changelog. This new docstring retells that two copies drifted and that a rejected write used to retry forever. The operator needs "write failed_state, then complete the row, even if the write fails."

**Suggestion:** Keep those two sentences. Delete the "used to" history.

django_logic/background/apps.py
```diff
@@ -1,4 +1,94 @@
...
+    conf.validate_core_settings()
+    if mode == conf.EXECUTION_PULL:
+        # The claim needs real row locks (SKIP LOCKED), and the state lock
+        # must span the web process and the worker processes.
+        _reject_sqlite_in_pull_mode()
+        _check_lock_cache_in_pull_mode()
+
+
+def _reject_sqlite_in_pull_mode() -> None:
+    """SQLite doesn't support ``select_for_update(nowait=True)`` nor
```
emil-balashov: **[suggestion]** The human asked whether SQLite support is needed. The code answer is fine: keep it for sync tests, refuse it in pull. The docstring is still a novel (partial unique indexes, secondary aliases, `override_settings`). That is the same over-explaining the comment was pointing at.

**Suggestion:** One sentence: pull mode needs SKIP LOCKED, so refuse SQLite on the TransitionMessage alias.

docs/history/BACKGROUND_TRANSITION_ANALYSIS.md
```diff
@@ -1,5 +1,11 @@
 # Background Transition — Design & Failure Analysis
 
+> **Historical record.** This file describes the broker design (Celery
```
emil-balashov, line 3: **[suggestion]** #222 said a notice is not enough because assistants will treat the file as current. This PR added the notice and left celery sections 6–9 in place. CLAUDE.md (not in this diff) still says to read this file before changing the engine, and still lists `dispatch.py` and `settings.py` as live modules.

**Suggestion:** Cut the celery sections, or move the file under a history path and point CLAUDE.md at `PULL_WORKERS.md` only. Fix the CLAUDE.md module list in the same pass.

django_logic/background/pull.py
```diff
@@ -218,12 +269,32 @@ def _wait_for_work(timeout: float) -> None:
...
-            cursor.execute(f'LISTEN {NOTIFY_CHANNEL}')
-        select.select([raw], [], [], timeout)
-        raw.poll()
-        raw.notifies.clear()
-    except Exception:
+        if not getattr(raw, '_django_logic_listening', False):
+            with raw.cursor() as cursor:
+                cursor.execute(f'LISTEN {NOTIFY_CHANNEL}')
+            raw._django_logic_listening = True
+        if hasattr(raw, 'poll'):
```
emil-balashov, line 356: **[suggestion]** #227 asked to delete LISTEN/NOTIFY and sleep the poll interval. The PR kept it and branched on `hasattr(raw, 'poll')` plus a `_django_logic_listening` flag on the raw connection. gv on psycopg 2 and a live NOTIFY test make a repair defensible on direct Postgres. This library's own deploy notes send workers through pgbouncer transaction pooling, which rejects LISTEN, so the dual wait becomes the except-and-sleep path anyway.

**Suggestion:** If pickup latency on direct Postgres is the reason to keep NOTIFY, say that in one sentence and keep one wait. If the documented deploy is pgbouncer, delete the wake-up as the issue asked.

django_logic/background/safety_nets.py
```diff
@@ -61,7 +62,7 @@ def run_pending(queues: list[str] | None = None) -> int:
     the runner's own row-lock guard.
     """
     ran = 0
-    for pk in list(_claimable(queues).values_list('pk', flat=True)):
+    for pk in list(_claimable(None).values_list('pk', flat=True)):
```
emil-balashov, line 65: how does it even work?

django_logic/background/pull.py
emil-balashov: Renamed in b560bcf: `_record_attempt_error(pk, message)` — the name states the work. Both callers and the two test references are updated.

django_logic/background/pull.py
```diff
@@ -102,61 +101,123 @@ def run_once(queues: list[str], *, isolate: bool = False) -> bool:
...
+            sent_kill = False
+            try:
+                os.kill(attempt_pid, signal.SIGKILL)
+                sent_kill = True
+            except ProcessLookupError:
+                pass
+            try:
+                _, raw_status = os.waitpid(attempt_pid, 0)
+            except ChildProcessError:
+                return 0, False
```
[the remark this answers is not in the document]
emil-balashov: Fixed in b560bcf. The `ChildProcessError` handler now honours the kill flag: when the kill was sent it returns `(-signal.SIGKILL, True)`, so the attempt is charged its `[timeout]` error and the retry is paced. The `WNOHANG` poll got the same guard — the same reaping environment could crash the worker through it. Two mocked tests pin both paths.

README.md
emil-balashov: Fixed in 7fe3932: "Two safety nets".

django_logic/background/runner.py
emil-balashov: Trimmed in 7fe3932. The docstring states the rule only: write failed_state, then complete the row even if the write fails — completing the row is what stops the retry loop.

django_logic/background/apps.py
emil-balashov: Cut to one sentence in 7fe3932: pull mode claims rows with SELECT FOR UPDATE SKIP LOCKED, so refuse SQLite on the alias that stores TransitionMessage.

docs/history/BACKGROUND_TRANSITION_ANALYSIS.md
emil-balashov: Done in 7fe3932. The celery-era analysis doc moved to docs/history/ with its links updated; CLAUDE.md's module list names the files that exist and points at PULL_WORKERS.md; the celery keyword left pyproject. docs/history/ is stated as the home of superseded design records.

django_logic/background/pull.py
emil-balashov: Decision: keep the wake-up and say why. pull.py's module docstring and PULL_WORKERS.md section 5 now say it exists for pickup latency on a direct Postgres connection (0.9s measured against the 5s poll floor) and falls back to the poll floor behind pgbouncer, which rejects LISTEN. The psycopg 2 branch also stays: gv pins psycopg2 < 3, so cutting it is a support-matrix change for whichever release declares psycopg 3.2+ for workers. #227 closes with this PR.

django_logic/background/models.py
emil-balashov: Agreed the branch delete was not the fix, and the maintainer has decided: one INSTALLED_APPS entry. The fold moves TransitionMessage into the django_logic app — one fresh initial migration under the new label, existing tables adopted with migrate --fake-initial — so the #233 squash arrives as a by-product instead of as surgery on applied migrations. #242 carries the concrete plan for review; until it lands, both entries stay required, and the README says so where it lists them.

django_logic/background/safety_nets.py
emil-balashov: Line 65 takes one snapshot of the primary keys of every row a worker could claim right now; the loop then runs each row inline.

`_claimable(None)` builds one SELECT over TransitionMessage. It keeps rows where `is_completed=False`, `errors_count < MAX_ERRORS`, and `last_error_dt` is null or older than now minus RETRY_MINUTES. Passing `None` means no queue filter, so rows from every queue qualify. `order_by('created')` returns them oldest first. `.values_list('pk', flat=True)` fetches only the primary keys, and `list(...)` runs that query once, before the loop starts — the loop walks a fixed snapshot and holds no open cursor across attempts.

For each pk, `run_background_transition(pk)` runs one attempt in this process. That call re-locks the row with `select_for_update(nowait=True)` and re-checks `is_completed=False`, so a stale pk — a row another connection completed mid-scan — exits silently as nothing-to-do. A failing row logs and the scan continues, so one bad row cannot stop the pass.

The filter is the same visibility rule the pull claim uses; `_claimable` is the one place it is written. So this pass runs exactly the rows a worker would take now, which is what makes `retry_pending()` an honest simulation of a worker pass.

django_logic/background/models.py
emil-balashov: Taking the point properly this time, and the answer is: two apps stay in 0.17.0, and the README now says so plainly instead of me presenting a branch delete as the fix.

The README's install section lists both entries as two entries, with no apology and no "this will change" note. #242 carries the fold with the concrete plan — `TransitionMessage` moves under the `django_logic` label, one fresh initial migration, existing tables adopted with `migrate --fake-initial`, and #233's squash arrives as a by-product rather than as surgery on applied migrations.

It is not in this release on purpose. gv went to pull mode in production this week and lost 18 minutes to a deploy-ordering trap. The next release it takes is the one that fixes that trap. A migration-identity change on the table holding live in-flight work does not belong in the same deploy — if that deploy goes wrong, nobody could tell which change did it.

One thing the consumer audit turned up that #242 should answer before it lands: gv already lists only `django_logic_new.background`, not the base app, and it has been working. The base `AppConfig.ready()` never runs there; the background app's ready hook does the boot validation. So the second entry the README asks for is already optional in practice — which raises the question of whether the fold is the fix, or whether making one app's `ready()` sufficient is. That is in the issue.

django_logic/background/pull.py
emil-balashov: The measurement is written down now, which it was not when I said it was: `docs/design/PULL_WORKERS.md` section 5 records pickup at 0.9 s against the 5 s poll floor, measured on the Heroku rig. The module docstring and the design doc both say the wake-up exists for a direct Postgres connection and falls back to the poll floor behind pgbouncer.

emil-balashov:
**The Heroku gate is green: 33 of 33, plus the two manual rows.** Merging.

Run recorded in `django-logic-test` at `docs/RESULTS.md`, engine @ `2a63d4b5`, on PostgreSQL + Redis with `worker_critical` at `--concurrency 4`.

Three things the gate itself turned up, all fixed before this run:

**The harness was reporting false green, and had been since its first commit.** A row's outcome was decided by grepping the validator's JSON for `"ok": true` — and that string also appears in every *passing sub-check*. So a scenario whose verdict was `false` reported PASS as soon as one of its checks passed; a row failed only when every check failed. I found it by insisting the new concurrency row prove it can fail: with a serial worker the validator answered `ok: false` and the row still said PASS. Every result recorded before this run is weaker evidence than it looked — the rows were fine, the reader of their verdicts was not.

**`--concurrency` did not work for the case it was built for.** The worker blocked in `waitpid` after every claim pass whether or not it had a free slot, so a burst arriving while one long attempt ran waited out that attempt: the notification wait was unreachable, and the poll floor with it. A full worker still blocks; one with a free slot now reaps what it can and goes back to the wait. The test that pins it fails in ten seconds against the old behaviour. The claim and the fork are also guarded now — a database blip used to end the loop, and with several attempts in flight that orphans them, so nothing would enforce their `timeout=` or record their crash.

**The harness's stall knob never stalled.** `stall_times` reached the harness API but neither the HTTP trigger nor the management command passed it on, so every earlier `timeout=` row ran without stalling. Wired in both places; TC-09 now exercises a real kill.

The new rows, both checked in both directions:

- **TC-35** — four attempts each sleeping 20 s, asserting two starts fall within 15 s. Passes at concurrency 4; **fails at concurrency 1** with a smallest gap of 20.1 s. The worker logs `attempts at a time=4` at boot, so the flag is confirmed on a real dyno.
- **TC-36** — `dl_transitions` names a row at `MAX_ERRORS`, and `--send` refuses both that row and a completed one.
- **TC-32** (manual) — a 120 s attempt reads as `an attempt has started it`, one row, `physical_runs=1`, lands `fulfilled` with 0 errors. No duplicate dispatch.
- **TC-34** (manual) — with no consumer on the queue, the listing says: `does a worker serve 'django_logic.slow'? Start one: dl_worker --queues django_logic.slow`. That is the line nobody could get on 2026-08-24.

TC-27 and TC-28 are retired with `DEFER_UNLOCK_UNTIL_COMMIT`. TC-26 had been skipping silently because its guard was coupled to that setting — it pins the lock-ownership contract again.

**One correction to this PR's own claim.** #245 unified the *lock* key, but `TransitionMessage` still keys the durable row on the stored model name, so a proxy pair writes two rows and the partial unique index does not tell them apart. My changelog said the index backstops the cache lock; that is true for every shape except the proxy pair #245 targets. The changelog now says so, and the row half is #258 — it changes instance identity on rows already written, so it needs a migration plan rather than a rushed edit.

**Consumer gate:** 2,212 gv tests pass against this head, re-vendored. Split into Borderless360/gv#9662 (the mirror) and Borderless360/gv#9661 (the six files that carry decisions).

Issue #260, opened 2026-08-24
Two dl_transitions messages promise a worker that may not exist


emil-balashov:
The MAX_ERRORS line says "the stuck finalizer ends it in its failed_state on its next pass", and `--send` ends with "A worker claims it next." The finalizer runs inside the worker loop, so with no worker on that queue neither pass ever comes. The listing's no-worker branch already says to start one, but only for rows that never started.

Make both lines conditional on a worker serving the queue, or append "when a worker serves <queue>" — one clause, no new mechanism.

PR #262, opened 2026-08-25
0.17.1 — the accounting write survives a database outage

### LINKED ISSUE #259: A database blip while the worker records a crash or timeout can kill the loop and lose the accounting
## Summary

`run_worker`'s exception guard covers only the claim/fork path. The `_harvest` calls (pull.py:433 and :441) sit outside it. `_harvest` pops the attempt from the dict, then `_account` runs `_record_attempt_error`, whose UPDATE has no guard. An `OperationalError` there propagates out of `run_worker` and the dl_worker process exits — with the attempt entry already popped, so the kill/crash record is lost.

The scenario is correlated, not exotic: a database outage makes attempts crash, the worker then records those crashes against the same dead database, raises, and dies while its other forked attempts still run. Nothing then enforces their `timeout=` or records how they ended. The follow-up commit that guarded the claim side names exactly this rationale; the harvest side does the same kind of database write and stayed unguarded.

## Suggested change

Guard the harvest accounting the same way the claim loop is guarded. Pop the attempt only after `_account` succeeds (or re-insert it when the write fails), so the next pass retries the write and the killed/exit information survives.

## Related observations, recorded but held under the release policy

- On `ChildProcessError` every remaining attempt is accounted as a clean exit (pull.py:219-224), so in an external-reaper environment a real crash gets an unpaced, uncounted retry. Needs a reaper in the worker parent, which no consumer runs.
- `waitpid(-1)` (pull.py:217) can steal the exit status of a subprocess a failure callback started, if that subprocess outlives the callback.

Found reviewing 0.17.0 as shipped, during gv's adoption review. Per the release policy this lands in the next consumer-driven release, not as an in-place fix.

emil-balashov:
## Summary

gv's adoption review of 0.17.0 found two defects, both in code 0.17.0 added. #261 landed the wording one (#260) with a more accurate condition than my draft had, so this carries the remaining one and the documentation corrections.

## A database blip while the worker records a crash or timeout no longer kills the loop (#259)

The claim and fork were guarded; the harvest's accounting write was not — and it writes to the same database whose outage crashes attempts in the first place. The failure is correlated, exactly as the issue says: attempts crash, the worker records the crashes against the dead database, raises, and dies while its other forked attempts still run. Nothing then enforces their `timeout=` or records how they ended.

The write cannot raise out of the worker loop now:

- A reaped attempt leaves the worker's books only when its accounting write lands. A failed write keeps it, marked `reaped`, and every later pass retries it — so the killed/crashed verdict survives the outage instead of being lost with the pop.
- When only failed writes remain there is no process to wait on, so the retry paces at one second rather than hammering the database.
- An unaccounted attempt keeps occupying its slot, so a worker that cannot record outcomes stops claiming new work until the database answers — backpressure during exactly the outage that caused it.

Two tests pin the retry, one per verdict: a crash stays `[crashed]` and a kill stays `[timeout]` across the failed write. Both were run against the unguarded code and error there.

The issue's two held observations stay held, per the release policy written on them.

## Documentation corrections

An accuracy audit of the shipped 0.17.0 against its own code found three misstatements, all fixed here:

- `docs/design/PULL_WORKERS.md` still promised a bespoke boot message for `'celery'`. The follow-up simplification pass removed it — an unknown mode now fails loudly naming the valid ones.
- The changelog credited the `DEFER_UNLOCK_UNTIL_COMMIT` removal with "651 net lines". The commit is 646 net, and that total includes the re-export removal that shared it.
- The re-export removal is a breaking change with an upgrade action, so it carries the **Upgrade step** label the `enable_sync()` change already had.

The same audit verified the whole README line by line against the package and found it fully accurate, and swept the entire `v0.16.0..v0.17.0` public surface for an undocumented break — there is none.

## Validation

- SQLite 594, PostgreSQL 313, all 14 CI jobs.
- **Heroku matrix 33 / 33** on the live rig, engine @ `3e364b45`, `worker_critical` at `--concurrency 4`. The first pass was 32/33 and the failure was the harness catching itself — three of its rows judged a remote command on its exit code alone, so any failure passed them. Details in the harness `docs/RESULTS.md` and the comment below.

Fixes #259

---

> [!NOTE]
> **Medium Risk**
> Changes pull-worker failure handling on the hot path where attempts are reaped and errors recorded; behavior under DB outages is intentionally different and affects worker liveness and backpressure.
> 
> **Overview**
> **0.17.1** patches a worker bug from 0.17.0: after a forked attempt ends, `_harvest` could raise on the crash/timeout **accounting** write—the same DB outage that killed the attempt—taking down the worker while other attempts kept running with no timeout enforcement.
> 
> Accounting is now **deferred and retried**: reaped attempts stay in the worker’s book with `reaped` / `exit_code` until `_try_account` succeeds; failures are logged and retried (1s pacing when only pending writes remain). Unaccounted attempts still occupy concurrency slots, so the worker stops claiming new work until the DB recovers.
> 
> Two harvest tests lock in retry behavior for `[crashed]` and `[timeout]`. Docs/changelog fixes: unknown `BACKGROUND_EXECUTION` boot behavior in `PULL_WORKERS.md`, line-count and **Upgrade step** notes for the 0.17.0 re-export removal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8318d94ccd2823aa4d4c036101e3debc7cf58b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 The two defects the consumer's adoption review of 0.17.0 found (#259,
 #260). Both sit in code 0.17.0 added, so the fix ships before any
-consumer runs it.
+consumer runs it. The wording fix came in through #261.
 
 ### Fixed
 
@@ -22,10 +22,11 @@ consumer runs it.
   pin the retry, one per verdict: a crash stays a crash and a kill
   stays a timeout.
 - **`dl_transitions` stops promising a worker that may not exist**
-  (#260). The finalizer it points at runs inside the worker loop, so
-  the MAX_ERRORS line now says the pass comes when a worker serves the
-  row's queue, and `--send` says the next worker serving that queue
-  claims the row.
+  (#260). Both lines pointed at work that only happens inside a worker
+  loop. The MAX_ERRORS line now says the finalizer's pass comes from
+  any running worker — that sweep reads every queue, not just the
+  worker's own — and `--send` says a worker claims the row once one
+  serves its queue, because the claim *is* queue-filtered.
 
 ## [0.17.0] — 2026-08-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 ## [Unreleased]
 
+## [0.17.1] — 2026-08-25
+
+The two defects the consumer's adoption review of 0.17.0 found (#259,
+#260). Both sit in code 0.17.0 added, so the fix ships before any
+consumer runs it.
+
+### Fixed
+
+- **A database blip while the worker records a crash or timeout no
+  longer kills the loop or loses the accounting** (#259). The claim and
+  fork were guarded; the harvest's accounting write was not, and it
+  writes to the same database whose outage crashes attempts in the
+  first place. The write now cannot raise out of the worker loop: a
+  reaped attempt leaves the worker's books only when its write lands,
+  and every later pass retries a failed one. An unaccounted attempt
+  keeps occupying its slot, so a worker that cannot record outcomes
+  stops claiming new work until the database answers again. Two tests
+  pin the retry, one per verdict: a crash stays a crash and a kill
+  stays a timeout.
+- **`dl_transitions` stops promising a worker that may not exist**
+  (#260). The finalizer it points at runs inside the worker loop, so
+  the MAX_ERRORS line now says the pass comes when a worker serves the
+  row's queue, and `--send` says the next worker serving that queue
+  claims the row.
+
 ## [0.17.0] — 2026-08-24
 
 The cleanup after the pull cut (#221): the words now match the engine,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,12 +176,13 @@ consumer API is unchanged except where noted.
   unchanged.
 - **`DEFER_UNLOCK_UNTIL_COMMIT` is removed** (#244). Nothing set it.
   The knob, its deferred-unlock registry, the `deferrable=` threading
-  through every release path, and its tests are gone — 651 net lines.
+  through every release path, and its tests are gone — 646 net lines,
+  counted with the re-export removal that shared the commit.
   Locks release when the transition finishes; drive follow-up work
   from `transaction.on_commit` when it must see the caller's commit.
   The key joins the removed-settings tombstones.
 - **The top-level re-export of the command classes is removed**
-  (#244). Import `Conditions`, `Permissions`, `SideEffects`,
+  (#244). **Upgrade step**: import `Conditions`, `Permissions`, `SideEffects`,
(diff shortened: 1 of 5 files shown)


emil-balashov:
**Heroku gate green: 33 of 33** on the live rig, engine @ `3e364b45`, `worker_critical` at `--concurrency 4`. Recorded in the harness `docs/RESULTS.md`.

The first pass came in 32/33, and the one failure was the harness catching itself. Three of its rows could pass without running: `expect_fail` judged a row on the exit code alone, so any failure passed it — a missing argument, an expired heroku session, a dyno that never booted. Both TC-36 `--send` rows used it, and both take an id parsed out of a listing, so an empty parse would have run the command with no value and reported PASS.

With the rule tightened to require the refusal *text*, TC-36b failed immediately and correctly: the stuck finalizer completes a MAX_ERRORS row about once a minute, and `--send` runs in its own dyno some twenty seconds after the listing, so the row was already completed and the command refused it as completed rather than as exhausted. Both refusals are right; which one lands is a race the row cannot win. It accepts either now and still rejects a success or an unrelated failure. Re-run: 5/5.

Also hardened: TC-35 could have passed on a serial worker (its event is written before the sleep, so an attempt dying in between frees a slot in under a second) — it now asserts every attempt ran exactly once and that one dyno ran them all.

Suites: SQLite 594, PostgreSQL 313, all 14 CI jobs, harness pytest 78. Merging.

emil-balashov:
Rebuilt on master after #261 landed the #260 fix. I dropped my version of it: yours states the condition more accurately. The stuck sweep in `detect_stuck_transitions` reads every queue, not just the running worker's own, so "the next pass of any running worker" is right where my draft said "when a worker serves this queue". The claim *is* queue-filtered, so naming the queue on `--send` is correct there. The changelog now describes what shipped.

**Heroku gate: 33 / 33** on the live rig, engine @ `3e364b45`, `worker_critical` at `--concurrency 4`.

The first pass came in 32/33, and the failure was the harness catching itself rather than the engine. Three of its rows judged a remote command on its exit code alone, so *any* failure passed them — a missing argument, an expired heroku session, a dyno that never booted. Both TC-36 `--send` rows used that check, and both take an id parsed out of a listing, so an empty parse would have run the command with no value and still reported PASS.

Tightened to require the refusal text, TC-36b failed at once and correctly: the stuck finalizer completes a MAX_ERRORS row about once a minute, and `--send` runs in its own dyno some twenty seconds after the listing, so the row was already completed and the command refused it as completed rather than as exhausted. Both refusals are right; which one lands is a race the row cannot win. It accepts either now and still rejects a success or an unrelated failure. Re-run 5/5.

Suites: SQLite 594, PostgreSQL 313, 14/14 CI, harness pytest 78.

Issue #263, opened 2026-08-25
run_once(isolate=True) discards a failed accounting write


emil-balashov:
In `_harvest`, the reap path ends with `if _try_account(attempt): del attempts[pid]` followed by an unconditional `if block: return`. With `block=True` it returns even when the write failed. `run_once` builds a local attempts dict, calls `_harvest(block=True)` once, and drops the dict — so the kept entry, and with it the crash or timeout error, is lost on that path. A crash loop driven through `run_once` is never bounded by `MAX_ERRORS`.

Scope: only tests call `run_once` today; `dl_worker` runs `run_worker`, whose dict persists across passes, so production retries the write as designed.

Fix options: loop `_harvest` in `run_once` until the dict empties, or state in `run_once`'s docstring that it does not retry a failed accounting write.

Found gate-reviewing 0.17.1 for gv's adoption. Filed per the release policy: findings from later passes become issues for the next consumer-driven release.

emil-balashov:
Fixed in 0.17.2 (PR #264): run_once loops the harvest until the accounting write lands.

PR #264, opened 2026-08-25
0.17.2 — a failed accounting write keeps its retry pace

emil-balashov:
Two fixes in the accounting-retry path 0.17.1 introduced, both found reviewing 0.17.1 during gv's adoption (the vendor PR's bot flagged the first; #263 records the second).

## What changed

- **A failed accounting write no longer waits for a sibling attempt to end.** When every slot was full and a live attempt carried no `timeout=`, the worker blocked in `waitpid` — so the retry of a failed crash or timeout write waited for that sibling's whole run, and the crashed row stayed unpaced for that long. The worker now polls at the one-second pace while a write is waiting. One condition in `_harvest`.
- **`run_once` returns only after the accounting write lands** (#263). It ran one harvest pass and dropped a failed write with its verdict. It now loops until the write lands. Only tests call `run_once` today.

Two new tests pin both: the retry lands while a three-second sibling still runs, and `run_once` retries a refused write to completion.

## Scope notes

- The reclaim-race half of the bot's finding needs a second worker on the same queues; gv's target formation is one dyno, and when it does fire the charged events are all real — `errors_count` stays honest, only the message/date ordering can skew. No further machinery for it.
- Commit e90bf98 (RELEASING.md) was authored on this branch by a concurrent session or by Emil directly while the fix was being written; it is doc-only and rides along.

## Validation

- SQLite suite: 596 tests OK
- PostgreSQL (tests.background + tests.stability): 315 tests OK

---

> [!NOTE]
> **Medium Risk**
> Changes core worker harvest/wait logic for multi-slot workers and failed DB accounting writes; behavior is pinned by new fork-based tests but affects crash/timeout error recording in production.
> 
> **Overview**
> **Release 0.17.2** tightens pull-worker accounting when a crash/timeout write fails or when several attempts run at once.
> 
> In **`_harvest`**, if a reaped attempt still needs a database write, the worker no longer blocks in **`waitpid`** while a sibling with no **`timeout=`** runs. It polls with **`WNOHANG`** so failed writes retry on the usual ~1s cadence instead of waiting for the sibling’s full run.
> 
> **`run_once`** now loops **`_harvest`** until the attempt’s accounting write succeeds, so a single harvest pass cannot return after dropping a refused write (today only tests call **`run_once`**).
> 
> Changelog and version bump to **0.17.2**; new tests cover sibling polling and **`run_once`** retry-to-completion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f29f30a125bdca35ab2df533865c727a2077fa2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## [Unreleased]
 
+## [0.17.2] — 2026-08-25
+
+### Fixed
+
+- **A failed accounting write no longer waits for a sibling attempt to
+  end.** When every slot was full and a live attempt carried no
+  ``timeout=``, the worker blocked in ``waitpid``, so the retry of a
+  failed crash or timeout write waited for that sibling's whole run.
+  The worker now polls while a write is waiting, and the retry keeps
+  its one-second pace. Found by review during gv's adoption of 0.17.1.
+- **``run_once`` returns only after the accounting write lands** (#263).
+  It ran one harvest pass and dropped a failed write with its verdict;
+  it now loops until the write lands. Only tests call ``run_once``
+  today, so no deployment saw the loss.
+
 ## [0.17.1] — 2026-08-25
 
 The two defects the consumer's adoption review of 0.17.0 found (#259,
--- a/django_logic/background/pull.py
+++ b/django_logic/background/pull.py
@@ -112,7 +112,8 @@ def run_once(queues: list[str], *, isolate: bool) -> bool:
     segmentation fault, the platform's memory killer — kills the attempt,
     not the worker. The worker records the crash as an error on the row,
     so a crashing attempt gets the same paced, bounded retries as a
-    failing one.
+    failing one. The call returns only after the accounting write lands,
+    so a caller never drops a failed write.
     """
     from django_logic.background.runner import run_background_transition
 
@@ -122,7 +123,8 @@ def run_once(queues: list[str], *, isolate: bool) -> bool:
     if isolate and hasattr(os, 'fork'):
         attempts: dict[int, _Attempt] = {}
         _start_attempt(pk, attempts)
-        _harvest(attempts, block=True)
+        while attempts:
+            _harvest(attempts, block=True)
     else:
         run_background_transition(pk)
     return True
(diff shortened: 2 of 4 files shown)


emil-balashov:
Reviewed and verified. Both fixes are real holes in the accounting-retry path I added in 0.17.1, and the change is correct.

**The two edits are coupled and both are needed.** Forcing `WNOHANG` while a failed write waits is the fix; the `next_deadline is None` branch on the sleep is what keeps it from crashing. Without that second edit the new poll path reaches `time.sleep(min(1.0, max(0.01, next_deadline - now)))` with `next_deadline` unset and raises a `TypeError` out of the worker loop — the exact class of failure 0.17.1 was closing. Worth noting in case anyone is tempted to take one edit without the other.

**Traced every branch of `_harvest` for spin and starvation:**
- Mixed live and failed-write attempts: `WNOHANG` returns `(0, 0)`, the sleep paces at one second, the loop top retries the write. No spin.
- Only failed writes left: the earlier `all(reaped)` branch already handles it and returns when `block=False`. The two conditions compose rather than overlap.
- `block=False` with a failed write and a live sibling: returns at once, and `run_worker` comes back through `_wait_for_work(BUSY_POLL_SECONDS)`. Still one second.

**Both new tests are genuine gates.** I reverted each edit and ran the matching test:
- Reverting the poll: `test_a_failed_write_retries_while_a_sibling_still_runs` FAILS (3.1 s — it waits out the sibling).
- Reverting the `run_once` loop: `test_run_once_retries_a_failed_accounting_write` FAILS with `1 != 2` — the write is attempted once and dropped.

**One thing to know rather than change:** `run_once` now loops until the write lands, so a permanently refused write makes it loop forever. That is right for the contract, and every caller is a test today. The tests use finite `side_effect` lists so none can hang. If `run_once` ever gets a production caller, that becomes a real consideration.

**Suites on the branch head:** SQLite 596 OK, PostgreSQL 315 OK. Matches your numbers.

Approving. This ships together with the gv adoption pair, so I will cut 0.17.2, gate it on the Heroku matrix, and re-vendor before merging gv#9662 and gv#9661.

Issue #258, opened 2026-08-24
TransitionMessage keys the row on the proxy's model name, so one row can carry two uncompleted transitions


emil-balashov:
## Summary

0.17.0 fixed the state lock so a proxy model and the model it proxies take one lock on one row (#245). The durable row was left keyed the old way, so the two halves now disagree.

`TransitionMessage.instance_key` (`background/models.py`) writes and queries with `instance._meta.model_name`. For a proxy that is the proxy's own name. So the partial unique index `dl_bg_one_uncompleted_per_process` — the constraint that enforces "one uncompleted background transition per instance per process" — cannot tell a proxy row from its concrete row.

## What that costs

Two background transitions on **one physical row and one state column**, enqueued one after the other through two proxies, both succeed:

- The first takes the state lock, writes its row, releases the lock.
- The second takes the same lock (now free), and `_enqueue_atomic` finds no unique conflict because `model_name` differs.

Both rows are uncompleted. A worker claims each. The side-effects run twice on one row. Reproduced with a `BackgroundAction`, so no state change hides it: the side-effect ran for the same pk twice.

The claim's new `exclude_pks` filter does not help — the two rows have different primary keys, so one worker with several slots can run both at the same time.

0.17.0's lock fix does close the **concurrent** race, because both enqueues now contend for one cache key. It does not close the sequential one, which is the common shape.

## Why this was not fixed in 0.17.0

Changing the key means writing the concrete model's name into new rows. Rows already in the table carry the proxy name, so on the deploy `in_flight_for` would stop matching them and the engine would read an in-flight transition as absent. `_restore` also loads the instance from `app_label` + `model_name`, so it would start restoring the concrete class where it used to restore the proxy.

That is instance identity on a durable row. It needs a migration plan for the existing rows and a rehearsal, which is the same gate as #242.

## Suggested change

Key the row on the concrete model, the way the lock key now names the declaring table. Decide first:

- what happens to uncompleted rows written under a proxy name at deploy time — a data migration, or a read that accepts either name for one release;
- whether `_restore` should load the concrete model or keep the recorded class, given `owning_process_class` already selects the transition.

## Consumer

gv's `export_app` binds four processes to one table through proxy models (`export_app/apps.py`). In practice a given row is created as one type, so this is latent there rather than active — but it is the exact shape #245 was filed for.

## Found in

The 0.17.0 review, after #245 landed.

PR #265, opened 2026-08-25
#242 — one INSTALLED_APPS entry

### LINKED ISSUE #242: One INSTALLED_APPS entry: the background engine is default
## Summary

Consumers still list `django_logic` and `django_logic.background`. The background engine is default functionality.

The background app owns the `django_logic_background` label and its migrations, so the fold must go with the migration squash (#233).

## Suggested change

One entry in `INSTALLED_APPS`. Do this now, with the squash. Do not wait for a 1.0 release.

The consumer still has to add the second app until this lands. That is not a reason to keep two apps.

emil-balashov:
## Summary

The README asked consumers for two entries. Only one was ever load-bearing, so the defect was the sentence and not the code. **Zero library code changes here.** No migration, no app label change, no table change.

`'django_logic.background'` already owns the `TransitionMessage` table and its nine migrations, the `dl_worker` and `dl_transitions` commands, and every system check — and its boot gate is a strict superset of the base app's. gv has run that single entry in production for months (`gv/settings/base.py` lists it and nothing else).

## What changed

- **README**: one entry, plus the two things a reader needs to know — an existing install can delete the old `'django_logic'` line whenever it likes, and `'django_logic'` **alone** is the sync-only shape, which is also one entry.
- **`tests/settings.py`**: installs the documented entry. It installed both, so 596 tests exercised a configuration no consumer had. The two tests that assert the base app's own `ready()` now install it for their own duration with `@modify_settings`, which turns the superset relation into something CI checks rather than an accident.
- **CHANGELOG**.

## Why the sync-only paragraph is not optional

Measured on Django 6.1, default pull mode, SQLite, per-process cache, `DEBUG=False`:

- `INSTALLED_APPS = ['django_logic']` → boots.
- `INSTALLED_APPS = ['django_logic.background']` → `ImproperlyConfigured` from the pull-mode boot gate.

A consumer running no background transitions needs the *other* single entry. Saying "one entry" without saying which would cost them their dev box.

## Why the fold is declined, not deferred

The plan on #242 was to move `TransitionMessage` into the `django_logic` app label. That was designed out in full and measured end to end — it works — and it is still the wrong trade:

- The app label is the address of the live table and of every row in `django_migrations`. Moving it buys a prettier `INSTALLED_APPS` line.
- It costs nine permanent `replaces` literals, a `db_table` pin that becomes load-bearing with no test defending it, doubled ledger rows on every fresh install, and a forced settings edit on every existing consumer under threat of a boot failure.
- It can silently re-route `TransitionMessage` for any consumer whose database router keys on `app_label == 'django_logic_background'` — a shape the system checks literally instruct consumers to build.

The release policy in `CLAUDE.md` bans machinery without a consumer-reproduced defect. Nobody has reported one caused by the label. The label is ugly, not broken.

**#233 is unaffected** and stays open as a normal same-label `squashmigrations` job, with no dependency on this.

## Validation

- SQLite 596 OK, PostgreSQL 315 OK — under the single documented entry.
- The sync-only shape verified booting on SQLite with the default cache.
- Both `@modify_settings` decorators verified load-bearing: without them those two tests fail with `LookupError: No installed app with label 'django_logic'`.

Fixes #242

---

> [!NOTE]
> **Low Risk**
> Documentation and test settings only; existing dual-entry installs remain supported with identical behavior.
> 
> **Overview**
> Documents **one** `INSTALLED_APPS` line for the full library: **`django_logic.background`** alone. The README and changelog state that entry already owns migrations, worker commands, and system checks, and that keeping the old **`django_logic`** line is harmless but optional.
> 
> **Sync-only** installs are documented separately: **`django_logic`** alone (no table, worker, or DB/cache boot checks).
> 
> **No runtime or migration changes** — app labels and tables stay as they are; folding `TransitionMessage` into the base app is explicitly out of scope.
> 
> The **test suite** now matches the documented install by dropping **`django_logic`** from `tests/settings.py`. Two tests that exercise the base app’s `ready()` hook temporarily append **`django_logic`** with `@modify_settings` so CI still verifies that path without running the whole suite under the old dual-entry setup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61b0e0066e9e92337dfb760ff86c58f3470b5963. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **One `INSTALLED_APPS` entry** (#242). The documented install is
+  `'django_logic.background'` alone. That entry always owned the
+  `TransitionMessage` table and its migrations, the `dl_worker` and
+  `dl_transitions` commands, and every system check, and its boot gate
+  is a strict superset of the base app's — the second entry was never
+  load-bearing. The consumer has run this shape in production for
+  months.
+- No code changed, no migration ran, and no app label or table name
+  moved. An install that keeps `'django_logic'` boots and behaves
+  identically, so delete that line whenever you like.
+- `'django_logic'` alone stays the sync-only shape — also one entry, and
+  the one to keep if you run no background transitions. It has no table,
+  no worker, and no boot check on the database or the cache.
+- The test suite now runs the documented shape. It installed both
+  entries, so 596 tests exercised a configuration no consumer had.
+- This does **not** fold `TransitionMessage` into the `django_logic` app
+  and does **not** squash the nine migrations (#233). Folding would move
+  the app label, which is the address of the live table and of every
+  row in `django_migrations`, to make one settings line prettier.
+
 ## [0.17.2] — 2026-08-25
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -33,15 +33,29 @@ pip install django-logic
 Add `[redis]` if your settings name `django_redis.cache.RedisCache`. That
 extra installs the third-party backend for you.
 
-Add two entries to `INSTALLED_APPS` and create the table:
+Add one entry to `INSTALLED_APPS` and create the table:
 
 ```python
 INSTALLED_APPS = [
     ...,
-    'django_logic',
     'django_logic.background',
 ]
 ```
+
+That one entry is the whole library. It owns the `TransitionMessage`
+table and its migrations, the `dl_worker` and `dl_transitions` commands,
+and every system check. `import django_logic` works as normal — a Python
+package needs no `INSTALLED_APPS` entry to be importable.
+
+Upgrading from a release that asked for a second entry, `'django_logic'`?
+Delete that line whenever you like. Keeping it is supported and changes
+nothing: it is a real app and both boot hooks are idempotent.
+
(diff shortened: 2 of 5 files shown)


emil-balashov:
Reviewed and verified independently before merge.

- SQLite suite under the single entry: 596 tests, OK.
- PostgreSQL suites on a local Postgres 17: 266 + 49 = 315, OK.
- `INSTALLED_APPS = ['django_logic']` alone boots on SQLite with the default cache and `DEBUG=False`.
- `INSTALLED_APPS = ['django_logic.background']` alone on SQLite refuses at boot with the pull-mode gate, as the README's sync-only paragraph says.
- The boot-gate superset claim holds in the code: the background app's `ready()` runs every step the base app's `ready()` runs, plus the background gates.

One correction to the body, for the record: the fold plan on #242 used one fresh initial migration adopted with `migrate --fake-initial`, not a `replaces` list. The other listed costs stand, and they are enough to decline the fold.

The declined fold also unhooks #258 and #233 from this change. #258 (the proxy row key) is a reproduced defect and ships next on its own. #233 stays a same-label squash.

Issue #266, opened 2026-08-25
Sibling proxies can bind one process name to two state fields; the row guard then gates them as one machine


emil-balashov:
## Summary

Since 0.18.0 the durable row keys on the concrete model plus the process name. Two sibling proxies of one model can each bind a process named `process` to a **different** state field: the duplicate-name check in `bind_model_process` compares `existing.model is model`, and sibling proxies are distinct classes, so both bindings are accepted.

Those are two independent state machines on one row. Before 0.18.0 each kept its own uncompleted row (the key held the proxy's name). Now they share one key, so the second machine's background transition raises `AlreadyInProgress` while the first machine's row is uncompleted, and a synchronous transition raises `TransitionTemporarilyUnavailable`.

## The remedy that exists today

Give the processes distinct process names. One class already must: `bind_model_process` refuses a duplicate name on the same class. The 0.18.0 changelog states the rule: the guard is per physical row and process name.

## Options if a consumer hits this

- Refuse at bind time: a proxy binding a process name its concrete model's other proxies already bind to a different field gets the same error one class gets today. Smallest, and it names the remedy in the message.
- Add `field_name` to the partial unique constraint and to the guard queries, so the machine — row plus field — is the unit. Bigger: the public probes take no field name.

## Found in

The one adversarial review pass of 0.18.0, reproduced on the branch with a script. Per the release policy it lands only when a consumer hits or reproduces it in a real configuration. gv's export_app binds its proxies to one field, so it does not hit this today.

emil-balashov:
Closing as not planned after the release scope review. Independent state fields need distinct process names under the existing per-physical-model/process contract. GV binds its export proxies to the same state field, so it does not need this conflicting configuration. Adding another identity contract or validator is not justified by current usage. PR #295 documents the unsupported binding boundary in the [README](https://github.com/Borderless360/django-logic/blob/a0531715ca4271cf0ad917f2cd1e9de2d72641c2/README.md). This closure does not claim that conflicting bindings were fixed or rejected automatically. Reopen when a concrete consumer needs this configuration.

PR #267, opened 2026-08-25
1.0.0 — one transition contract, the concrete row key, and the bridges removed

### LINKED ISSUE #233: Squash the nine TransitionMessage migrations now

### LINKED ISSUE #243: One transition type, one lock/gate/chain contract
## Summary

Four types pretend to be a 2x2. They are not.

A sync `Action` takes no lock, skips the in-flight gate, and ignores `next_transition`. `BackgroundAction` subclasses `BackgroundTransition`, so it does all three. The README table exists because the names lie.

## Suggested change

One class. `target=None` means no success write. Keep the durable variant as the smaller break for consumers. Document the migration in CHANGELOG.md.

Do this now. Do not wait for a 1.0 release.

### LINKED ISSUE #258: TransitionMessage keys the row on the proxy's model name, so one row can carry two uncompleted transitions

emil-balashov:
## Summary

This is the 1.0.0 release. Nobody ever ran 0.18.0, so the whole line ships as one release with one migration guide.

**One transition contract (#243).** `Action` and `BackgroundAction` are removed. `target=None` (the default) declares a transition that writes no state on success, on `Transition` and `BackgroundTransition` alike. Every declaration takes the state lock, is refused with `TransitionTemporarilyUnavailable` while a background transition is uncompleted, and runs `next_transition`. Importing the removed names raises `ImportError` carrying the migration.

**The proxy row key (#258).** The key columns name the concrete model, the way the state lock has since 0.17.0; the driving proxy class moves to the new `proxy_model_label` column and the worker restores it, so side-effects still receive the proxy instance. Migration `0010` rewrites uncompleted proxy-keyed rows and reverses cleanly.

**The squash (#233).** One initial migration for fresh installs; existing installs keep using the ten originals, which stay in the package until every install has applied the squash.

**The 1.0 bridge removal.** The maintainer waived the consumer-side gates and will run the gv port after the release:
- `STRICT_HOOK_SIGNATURES` / `STRICT_KWARGS_SERIALIZATION` are gone — strict is the only behavior. A bad hook signature fails at bind time (the `W001` re-report check is retired with it); a caller-passed `request` or `user_id`, and non-string dict keys, fail at enqueue. The chain hop strips `request` and `user_id` before a background follow-up, so both stay legal on the synchronous transition that chains.
- `LEGACY_EXCEPTION_BASE` and its boot-time base-class mutation are gone. All three removed keys carry `django_logic.W003` migration advice.
- `DjangoLogicException` and `SourceStateChanged` collapse into the exceptions consumers catch: `TransitionNotAllowed`, `TransitionTemporarilyUnavailable`, `AlreadyInProgress`, `PermanentFailure`.
- The `conditions_class` / `failure_callbacks_class` swap hooks and the empty `[celery]` extra are gone. `nested_processes` are frozen.
- Version 1.0.0, classifier Production/Stable.

## Validation

- SQLite: 590 tests OK. PostgreSQL: 279 + 49 OK. Voice and metadata checks OK. `makemigrations --check` clean.
- Migration rehearsal on PostgreSQL 17: upgrade path (nine applied → only `0010` runs, seeded proxy row rewritten, squash recorded) and fresh path (squash only) produce byte-identical schemas.
- The proxy-key tests fail 5-for-5 on the old keying; the old-contract pins were rewritten to pin the new contract and fail without it.

## The adversarial review passes

One pass per commit set, findings adversarially verified. They produced: the snapshot round-trip and operator-identity fixes, the `0010` reverse, the `waits`-vs-refuses wording corrections, and — in the bridge-removal pass — one real behavior catch: the chain hop stripped `request` but not `user_id`, so a legal sync kwarg silently killed a chain; it now strips both, with a pin. One narrow finding stays open per the release policy as #266.

## Deliberately still in the package

The ten replaced migration files (they leave once every install has applied the squash) and the `ImportError` guidance for the removed class names (the migration guide for the jump to 1.0).

Fixes #258
Fixes #233
Fixes #243

---

> [!NOTE]
> **High Risk**
> Major breaking API and semantics (removed Action types, stricter sync/no-target transitions, enqueue kwargs) plus a data migration on live `TransitionMessage` rows; proxy/concurrent-row edge cases require coordinated deploy and consumer migration.
> 
> **Overview**
> **django-logic 1.0.0** collapses four transition types into one contract and fixes proxy-model background rows sharing one physical instance.
> 
> **One transition contract (#243).** `Action` and `BackgroundAction` are removed; use `Transition` / `BackgroundTransition` with **`target=None`** for side-effects with no success state write. Every declaration now takes the state lock, blocks while an uncompleted background row exists (`TransitionTemporarilyUnavailable`), and runs **`next_transition`** — including former `Action` call sites that must be reviewed. Imports of removed names raise **`ImportError`** with migration text.
> 
> **Concrete row key for proxies (#258).** `TransitionMessage` uniqueness and lookups key on the **concrete** model; **`proxy_model_label`** records the class the caller drove and **`driving_model_label`** feeds operators/Sentry/`dl_transitions`. The worker restores the proxy instance for hooks. Migration **`0010`** rewrites uncompleted proxy-keyed rows (with clash handling); **`0001_squashed_0010_proxy_model_label`** gives fresh installs a single migration (#233).
> 
> **Strict-only behavior and bridge removal.** `STRICT_HOOK_SIGNATURES` / `STRICT_KWARGS_SERIALIZATION` are gone — bad hook signatures fail at **bind** (`W001` retired); `request`, caller `user_id`, and non-string dict keys fail at **enqueue** (`KwargsSerializationError`). `next_transition` into background strips **`request` and `user_id`** at the chain boundary. `LEGACY_EXCEPTION_BASE`, `DjangoLogicException`, and `SourceStateChanged` are removed; enqueue source-state races raise **`TransitionTemporarilyUnavailable`**. Swap hooks **`conditions_class`** / **`failure_callbacks_class`**, the empty **`[celery]`** extra, and related settings readers/boot hooks are dropped. Docs, tests, and **`pyproject.toml`** (1.0.0, Production/Stable) align with the new surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0109188436165bff8921487a63738420f49049d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


django_logic/background/runner.py
```diff
@@ -778,13 +778,16 @@ def _state_guard_matches(transition, state) -> tuple[bool, str, str]:
 
 def _restore(transition_message: TransitionMessage):
     """Resolve ``(instance, process, transition)`` from a TransitionMessage row."""
+    # The key columns name the concrete model; driving_model_label names
+    # the class the caller drove. Restore that class, so proxy methods and
+    # overrides stay visible to side-effects and callbacks. Rows written
+    # before proxy_model_label existed record the driving class in
+    # model_name, so the same read restores them unchanged.
+    recorded_model = transition_message.driving_model_label
```
emil-balashov, line 786: confusing name `driving model` when it's actually  recorded model

emil-balashov: Agreed. Renamed to `recorded_model_label`, and the whole prose family moved with it ("the class the caller drove" → "the class that recorded the row") — swept across models, runner, observability, dl_transitions, the 0010 migration prose and the snapshot helper. Shipped in 1.2.0 (PR #270); the `proxy_model_label` column and the schema are unchanged.

PR #274, opened 2026-08-27
2.1.0 — the diet: one way to test, no redis extra, one settings warning

emil-balashov:
## Summary

The three approved cuts from yesterday's audit, all evidence-backed:

**The `[redis]` extra is gone.** The engine's whole cache surface is `cache.add/get/delete` through Django's cache API — it imports no backend, and Django's own `RedisCache` satisfies it. Nobody installs the extra (gv vendors and pins django-redis itself; the harness pins it directly; CI never installed it), and pip *warns* on an unknown extra rather than failing, so stale pins keep installing the base package. The README loses the extra sentence and keeps the required cache contract; the E005 hint names "any shared backend" instead.

**One way to test.** `JourneyStep` and `assert_journey` are removed: the recap way could express nothing the plain asserts cannot (its only two unique powers — a closed ordered hook list and a post-hoc recap — are moot in real usage), while plain asserts express what it cannot (exception type at the caller, chain intermediates, field deltas, row facts). All five upstream recap tests were restatements of siblings and are deleted, not rewritten. `assert_state_trace` **stays** — it is the only way to pin the intermediate states of a chained drive, and its fifteen call sites keep that coverage. `record_driven_transitions`/`DrivenTransitions` leave with zero consumers in three repos.

**One settings warning.** `W003` folds into `W004`: a key a past release removed reports exactly like a typo — the value has no effect and the default applies — and the changelog owns each removal's upgrade advice. The eleven-key advice table (~109 lines with its tests) leaves with the id.

## Consumer notes

gv acquires the deletions on its next re-vendor; its follow-through is deleting ~211 trailing `assert_journey` recap lines across 14 test files (43 of 45 methods already carry full plain-assert coverage — deletion, not rewriting). The harness uses none of the removed surface beyond two comments.

## Validation

SQLite 580 OK, PostgreSQL 275 + 49 OK, voice OK.

---

> [!NOTE]
> **Medium Risk**
> Breaking for consumers silencing old check ids, pinning `django-logic[redis]`, or using removed test/worker flags; core transition and worker behavior changes are mostly dead-code removal and clearer failure-note semantics.
> 
> **Overview**
> **2.1.0** trims surface area and renames deployment checks so `SILENCED_SYSTEM_CHECKS` and docs use plain-word ids (`pull_mode_needs_postgresql`, `pull_mode_needs_a_shared_cache`, `transition_message_routing`, `unread_setting`) instead of `E002`/`E004`/`E005`/`W003`/`W004`.
> 
> **Packaging & ops:** The `[redis]` optional dependency is removed (lock stays Django cache API only); README/install hints point at any shared backend. `dl_worker` always runs the loop (`--once` gone). `run_worker` is no longer re-exported from `django_logic.background`. Tag publish is documented to build PyPI artifacts and create the GitHub release from the changelog section.
> 
> **Runtime simplifications:** Unknown `DJANGO_LOGIC` keys (including formerly “removed” keys) emit a single `unread_setting` warning—no per-key migration table in checks. Background kwargs serialization no longer rejects `request` (caller-level refusal since 2.0). Restore drops the blank `field_name` guard; failure finalization notes **replace** the column instead of appending. `Transition` no longer trips on removed `failure_side_effects` / `lock_timeout` kwargs.
> 
> **Testing API:** `JourneyStep`, `assert_journey`, `record_driven_transitions`, and `DrivenTransitions` are removed; `assert_state_trace` and existing plain asserts remain. Upstream recap tests are deleted; a small `expect_raises` mismatch guard is added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2219b72a8b8a7d4bb004652864d82baa3356fece. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


django_logic/testing/scenario.py
```diff
@@ -313,16 +276,3 @@ def _finish(self, label, instance, tracker, raised, before, *, action=None,
                 f'failure scenario this test intends.{elsewhere}',
                 instance=instance)
         self._record(label, 'OK' if raised is None else 'FAILED(injected)', detail)
-        # Record the journey step: the observable transformation this drive
```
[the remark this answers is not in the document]
emil-balashov: Real catch — the deletion span swallowed `_exc_names`, which the expect_raises mismatch path still calls; a wrong raise would have died in NameError. Restored, and the path is now pinned: a ValueError against expect_raises=KeyError fails naming both types.

tests/test_removed_settings_check.py
```diff
@@ -1,67 +1,34 @@
-"""Removed DJANGO_LOGIC keys are reported as W003, typos as W004.
+"""Every DJANGO_LOGIC key the engine never reads is reported as W004.
```
emil-balashov: W004 should be fully replaced by plain english across the project

emil-balashov: as any other non-readable shorts

emil-balashov: Done across the project. The four ids are plain words now: `django_logic.transition_message_routing` (was E002), `django_logic.pull_mode_needs_postgresql` (E004), `django_logic.pull_mode_needs_a_shared_cache` (E005), `django_logic.unread_setting` (W004) — code, tests, docs and CLAUDE.md all follow; the changelog records the mapping. No known install silences the old names, so the rename is free. gv's vendored copy picks it up on the next re-vendor (its docs reference the two pull-mode ids — that rides the same PR).

Issue #284, opened 2026-09-14
Fix the dead LISTEN/NOTIFY wake-up on psycopg2


Andrey-Omelyanuk:
## Summary

`_wait_for_work` caches its "already listening" flag on the raw driver connection. On psycopg2 that object is a C type with no `__dict__`, so the assignment raises `AttributeError`. The broad `except Exception` below catches it, and the wait degrades to `time.sleep(timeout)`. The LISTEN/NOTIFY wake-up has therefore never run on psycopg2.

This is not an outage: the poll floor still picks the work up within `POLL_SECONDS`. The cost is pickup latency, one redundant `LISTEN` round trip per wait, and one warning per wait.

## Affected code

`django_logic/background/pull.py:417-420` and `:438-443` (master e129f20).

```python
raw = connection.connection
if not getattr(raw, '_django_logic_listening', False):
    with raw.cursor() as cursor:
        cursor.execute(f'LISTEN {NOTIFY_CHANNEL}')
    raw._django_logic_listening = True   # AttributeError on psycopg2
```

Because the flag never sticks, the guard never closes. `LISTEN` is re-issued on every call, and the psycopg2 branch below it — including `del raw.notifies[:]` — is never reached.

## Scenario

1. Run against PostgreSQL with psycopg2. Django uses psycopg2 when psycopg 3 is absent.
2. Call `_wait_for_work` twice.
3. Every call logs:

```
pull: the notification wait failed ('psycopg2.extensions.connection' object has
no attribute '_django_logic_listening' and no __dict__ for setting new
attributes); sleeping the poll interval instead.
```

`tests/background/test_pull_workers.py::NotificationWakeUpTests` already reproduces this. On psycopg2 it fails:

```
AssertionError: 10.000947660999373 not less than 5.0
```

The wait takes the whole timeout instead of returning on the notification.

The suite does not catch it today. The `dev` extra installs `psycopg[binary]`, and psycopg 3's connection is a Python object that accepts the attribute. The flag sticks there, so the test passes. Uninstall psycopg 3 and the same test fails.

A consumer runs this in production. gv logs the warning about 20,000 times a day, without a break since 2026-08-25.

## Suggested fix

1. Store the flag on the Django connection wrapper, which accepts attributes. Key it on the raw connection object, so a reconnect issues `LISTEN` again on the new session:

```python
if getattr(connection, '_django_logic_listening', None) is not raw:
    with raw.cursor() as cursor:
[...]

Issue #283, opened 2026-09-10
Restore a lock-free, side-effect-only transition (the removed Action)


intr00der:
## Problem

`Action` was removed in 1.0.0. It was the only way to declare a step that runs a
side effect **without** taking the per-instance state lock. Its suggested
replacement — a `Transition` with no `target` — writes no state on success but
still takes the lock and is still subject to the background-in-progress gate.

So there is now no way to express a side-effect-only step that must not serialise
on the instance lock. Any such step declared as a `Transition` takes the
per-instance lock even though it changes no state.

## Impact (observed in production, downstream in gv)

Several side-effect-only steps were `Action`s and became `Transition`s on the
port. They are `sources=all_states` with no `target` — they change no state — but
each now takes the lock on a **shared** instance (the integration row), so
unrelated units serialise on one non-blocking lock and the losers fail instantly
with `TransitionNotAllowed("State is locked")`:

- The sales-channel tracking write-back. The lock is keyed on the integration, so
  one lock is shared by every fulfilment of a store; a dispatch batch whose
  shipments coincided within a second lost the lock and stalled — hundreds of
  orders sat unsynced in the store, recoverable only by a 24h sweep. Worked
  around by calling the side effect as a plain function instead of a transition.
- Other per-integration self-loop steps (tracking-history update, stock
  transaction creation) still do this — ~700–1800 `Lock failed …-status-<id>`
  per day across integrations.

The workaround (move the step out of the process into a plain function) works but
loses the process's declaration and the callbacks/failure hooks a transition
provides — so the logic that records success/failure has to be re-wired by hand
at every call site.

## Ask

Give the engine a first-class way to express a lock-free, side-effect-only
transition again. Either:

1. Restore an `Action`-like construct — side effects only, no `target`, **no
   state lock**, no background-in-progress gate; still runs `callbacks` /
   `failure_callbacks` so consumers don't have to re-implement them; or
2. A flag on `Transition` (e.g. `lock_free=True`, only valid with `target=None`)
   so a side effect that changes no state can opt out of the per-instance lock.

Either keeps side-effect-only steps inside the process without making unrelated
units contend on one instance lock.

## Note

The `Action` removal note says a side effect that must not obey the transition
contract "belongs in a plain method, not in the process." This issue is a request
to reconsider that: a plain method loses the process's declaration and the
success/failure/history callbacks, and the contention comes not from the mere
existence of a lock but from the lock being keyed on a **shared** instance for a
step that changes no state.

---
[...]

Issue #281, opened 2026-09-10
Decide whether side-effect-only transitions need a lock-free path after the 1.0.0 contract merge


Andrey-Omelyanuk:
## Summary

1.0.0 merged `Action` into `Transition(target=None)` (#243, #226). The stated rationale was
that the four types "pretend to be a 2x2" and the names lie — `Action`'s exemptions were
treated as an inconsistency to remove, and the break as documentation-only ("Keep the
durable variant as the smaller break for consumers. Document the migration in
CHANGELOG.md").

For one call shape those exemptions turned out to be load-bearing rather than accidental,
and removing them is **silently lossy at the caller**. This issue documents the full
behaviour delta and asks the design question it raises: **should the library offer a
backward-compatible path for a side-effect-only transition that must not take the instance
lock — or is "then it is not a transition, write it as a plain method" the final answer?**

The question is not hypothetical. The rename cost a production incident in gv
(Borderless360/gv#9892): more than 1,000 customer order write-backs left undelivered to
external stores past 24 h, still recurring for some channels two days after a partial fix.
Nothing about that was a bug in this library — every changed behaviour is the documented
1.0.0 contract, and the release notes named three of the four. It is a question of whether a
pure call-site rename should be able to acquire them at all.

## The behaviour delta in full

`Action` (≤ 0.17.x) vs `Transition(target=None)` (≥ 1.0.0). Line refs against 2.1.0
`django_logic/transition.py`:

| Aspect | `Action` | `Transition(target=None)` | Line |
|---|---|---|---|
| State lock on the success path | **never acquired** | `state.lock()` is the first act; `TransitionNotAllowed("State is locked")` when not acquired | 233 |
| Source revalidation under the lock | none | `_ensure_db_state_in_sources` | 258 |
| Background-in-progress gate | **none** — explicitly permitted to run while a background transition was in progress on the same instance + process | `_ensure_no_background_in_flight` → `TransitionTemporarilyUnavailable` | 259 |
| `next_transition` on success | not executed | executed | 306 |
| Failure path | lock taken **only** around the `failed_state` write, and that write skipped while an uncompleted background row existed (the worker owned the state field) | holds the success-path lock; `write_failed_state` then `_release_lock` in `finally` | 308–325 |
| State write on success | none | none — unchanged |

Four behaviours are acquired, and the call-site change that acquires them is
`Action(` → `Transition(`. In gv that was 24 process files renamed mechanically.

[...]

Issue #280, opened 2026-09-07
Decide the shape of the worker and operator tools: Emil asked to remove the dl_worker / dl_transitions management commands


emil-balashov:
## Why

Reviewing gv's 2026-09-07 release (Borderless360/gv#9793), Emil commented *"remove commands"* on the two vendored management commands, `django_logic_new/management/commands/dl_worker.py` and `dl_transitions.py`. gv's rule is that engine changes are made here first and re-vendored, so the decision belongs to this repo.

## The constraint

- `dl_worker` is the pull engine's worker: on Heroku it is the `dl_worker` Procfile process (`python manage.py dl_worker --queues ... --concurrency=${DL_CONCURRENCY:-1}`). Nothing else claims and runs background transitions. Removing it as it stands stops all background work.
- `dl_transitions` is the operator's view of uncompleted rows and the one lever to wake a waiting row (`--send PK`). It is the tool the release checklist tells the deployer to run after a release.

## The question for this issue

What shape should the worker and the operator tool take if not Django management commands? Options seen so far:

1. Keep `dl_worker` as the process entry point but expose it as a console script of the package (`django-logic worker ...`) so the Procfile does not go through `manage.py`; move `dl_transitions` into a Django admin page (list uncompleted rows, one action to wake a row).
2. Run the worker inside the existing Celery worker process (a Celery bootstep or a periodic claim task) so no separate process type exists; keep an admin page for the rows.
3. Keep both commands and close this as "won't do", with the reasoning recorded.

Emil to say which he meant; the gv vendoring follows once a version ships.

## References

- Borderless360/gv#9793 (review threads on the two files), gv `Procfile`, `docs/django_logic_new/README.md` §(2) and §(3).

emil-balashov:
Decision (Emil, 2026-09-16): keep both commands while they are helpful, and both are.

`dl_worker` is the pull engine's only worker entry point — the `dl_worker` Procfile process on Heroku. Nothing else claims and runs background transitions, so removing it stops all background work. `dl_transitions` is the operator's only view of uncompleted rows and the one lever to wake a waiting one (`--send <pk>`); it exists because during the 2026-08-24 incident no command could list the waiting rows or say why they waited, and the release checklist tells the deployer to run it after a release.

Option 3 from the issue, then: the commands stay as Django management commands. The review comment that opened this asked to remove them from the vendored copy in gv; the answer is that they are part of the engine and gv keeps the mirror whole.

PR #286, opened 2026-09-16
2.2.0: a transition may take no lock

emil-balashov:
## Why

1.0.0 folded `Action` into `Transition(target=None)`. The fold gave every former `Action` the per-instance state lock, and that lock is keyed on the bound row — not on what the side effect works on. A step declared on a parent and run once per child therefore serialised the children on one non-blocking lock. In gv a store's tracking post ran once per parcel; a dispatch batch lost the lock and the parcels' tracking never reached the store (Borderless360/gv#9892, #281, #282).

## What

**`Transition(..., lock=False)`** — a side-effect transition that takes no state lock and is not refused while a background transition on the instance is uncompleted. It keeps `conditions`, `permissions`, `side_effects`, `callbacks`, `failure_callbacks` and `next_transition`. It refuses a `target` and a `failed_state` (a state write must serialise on that state) and `BackgroundTransition` refuses it (enqueue writes the busy state and the durable row under the lock). It releases nothing on completion — `State.unlock()` with no token of its own deletes the key, which would free a lock a concurrent transition holds. `get_available_actions()` lists it while the row is locked, because it can be called.

**`Transition` refuses a keyword it does not read.** On 2.1.0, `lock=False` was accepted and did nothing — the lock was still taken. Verified before writing this: `Transition('demo', sources=['a'], lock=False)` constructed, stored no attribute, and `change_state` locked anyway.

**The tracker items that fit a 2.2.0:** `StrictPermissions` / `NoUserPermissions` (#276), `failed_state` refused on a synchronous transition with no side effects (#279), the testing guide on `expect_raises` for background drives (#278) and on the caller's object after a background drive (#277 — the refresh the issue proposed was tried and rejected: it made gv's import API tests report the follow-up's final state where production reports `validating`, so the guide says to `refresh_from_db()` or use `assert_state`). #285 (the psycopg2 LISTEN/NOTIFY fix, #284) is merged and rides in this release.

**Left open, on purpose:** #275 (a reason on `TransitionNotAllowed`) is the other half of the gv incident — the refusal was swallowed because "locked" and "forbidden" are one class — but it touches every raise site and deserves its own release. #280 (the shape of `dl_worker` / `dl_transitions`) needs Emil's pick among its three options. #266 / #268 land only when a consumer hits them, per the release policy. #283 is a duplicate of #282 and closes with it.

## Gates

- SQLite suite: 617 OK. Lock and permission tests on a real Redis cache: 35 OK.
- `tests.background` on PostgreSQL, as the stability job runs it: 282 OK.
- `tests.stability` on PostgreSQL + Redis: 52 OK — including a new test that runs ten `lock=False` callers on one row under real threads (zero refusals) against the lock-taking sibling under the same load (refusals).
- Consumer: gv's full suite against the vendored copy of this branch — 8,846 tests, the two failures reproduce on gv `staging` without this change. An AST scan of gv's 782 transition declarations found no keyword the new refusal rejects.
- The metadata drift check passes; the changelog header matches `pyproject.toml`.

---

> [!NOTE]
> **Medium Risk**
> Core FSM locking and availability rules change for opt-in lock-free transitions; stricter declaration validation may fail misconfigured processes at import/bind time.
> 
> **Overview**
> **Release 2.2.0** adds **`Transition(..., lock=False)`** for side-effect-only steps whose work is not the bound row (e.g. posting tracking per parcel on a store). Those transitions skip the per-instance state lock and the “background in flight” gate, can overlap on one instance, still honor conditions/permissions/hooks/chaining, and stay visible in **`get_available_actions()`** while the row is locked. They cannot declare **`target`** or **`failed_state`**, and **`BackgroundTransition`** rejects **`lock=False`** at declaration time.
> 
> The release also ships **`StrictPermissions`** / **`NoUserPermissions`** in **`django_logic.commands`** so nested processes can split one action name by caller (user vs automation) without ambiguous resolution.
> 
> **Declaration-time validation tightens:** unknown transition kwargs raise **`ImproperlyConfigured`** (so a previously ignored **`lock=False`** on 2.1.0 is now real), sync transitions with no side effects cannot declare **`failed_state`**, and docs/testing guidance clarify background-test semantics for **`expect_raises`** and enqueue-time state on the caller’s object.
> 
> Tests add concurrency coverage for lock-free vs lock-taking siblings; version is bumped to **2.2.0** in **`pyproject.toml`** with a matching changelog entry (including the psycopg2 LISTEN/NOTIFY worker wake-up fix noted there).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29750a25a89c6743045241fb331b53879af8c317. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


emil-balashov:
Heroku rig after the release: full matrix on 2.2.0, **48 / 48**, on psycopg2 (the rig switched drivers for #284). New rows TC-42/43 (`lock=False` under real concurrent callers, under a held lock, and beside an in-flight background row) and TC-44 (pickup inside 2s of the poll floor, no failed notification wait in the worker log). Details in django-logic-test `docs/RESULTS.md`.

PR #291, opened 2026-09-22
Keep worker deadlines responsive and reuse completed slots promptly

### LINKED ISSUE #287: Crash accounting waits on a reclaimed row and blocks sibling timeout enforcement
In django-logic **2.2.0** (`a74391ca19dc3fc6719b31d6615ef521b2ceb7b2`), recording a crashed attempt can block the worker supervisor behind a replacement attempt's PostgreSQL row lock. Other children then run beyond their declared timeouts.

The blocking write is [`django_logic/background/pull.py:364`](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/django_logic/background/pull.py#L364-L371). `_harvest` executes this accounting synchronously before it returns to checking sibling deadlines. Filtering by `is_completed=False` protects a completed row; it does not make the UPDATE nonblocking.

### Reproduction

1. An attempt crashes and releases its durable row lock.
2. Another worker claims that row and holds its lock while running the replacement attempt.
3. The original supervisor records the crash while it also supervises an unrelated child with `timeout=1`.
4. The accounting UPDATE waits for the replacement. The supervisor cannot enforce the sibling deadline during that wait.

Local reproduction used a real PostgreSQL transaction holding the replacement row for 3.5 seconds and a real forked sibling. Synthetic supervisor bookkeeping represents the already-reaped crashed child; database behavior is not mocked.

```text
first_harvest_elapsed=3.507s deadline=1.000s child_still_alive=True
child_finally_stopped_at=3.511s
error_count_recorded=1
```

Runtime: Python 3.12.13, Django 5.2.17, psycopg 3.3.6, PostgreSQL 17.9. This is a reproduced supported workload, not evidence of a specific production incident. A dedicated Heroku consumer-harness reproduction will be recorded before implementation changes.

### Expected behavior and regression gate

Crash accounting must defer without blocking the supervisor when another worker owns the row. The sibling must still be stopped within its timeout plus the worker's documented polling allowance. Once the row becomes available, record the crash once, or skip it if the replacement completed first. Assert child exit/deadline, replacement contention, error count, and worker survival; fixture/setup failure must not count as a reproduction.

Keep the completed-row guard and atomic increment. Prefer a nonblocking accounting acquisition using the existing pending-write state. This differs from #259 and #263: an ordinary lock wait raises no database exception, so retaining accounting after exceptions does not address it.

No library fix is included. First preserve the failing server regression, then fix upstream and rerun the same case against the release candidate.

### LINKED ISSUE #288: Stuck-job failure callbacks run in the supervisor and defeat sibling timeouts
In django-logic **2.2.0** (`a74391ca19dc3fc6719b31d6615ef521b2ceb7b2`), stuck finalization executes consumer failure callbacks inside the pull-worker supervisor. A blocking callback stops claims and deadline enforcement for the supervisor's other children.

[`run_worker`](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/django_logic/background/pull.py#L511-L513) invokes safety nets from its main loop. [`finalize_stuck_attempt`](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/django_logic/background/runner.py#L218-L220) then runs failure callbacks synchronously after committing the terminal state. These hooks do not run in the isolated attempt process, and the transition timeout does not bound them.

### Reproduction

1. Create an exhausted uncompleted row with a failure callback that sleeps for 3.4 seconds.
2. Start another queued attempt with `timeout=1`. Its side effect records a durable marker after 2.6 seconds, then continues waiting.
3. Run the actual pull-worker loop with concurrency two.
4. The safety-net callback runs in the supervisor PID. The supposedly one-second attempt remains alive long enough to write the marker.

Local PostgreSQL reproduction, using the real worker loop and forked attempt:

```text
failure_callback_pid_is_supervisor=True
worker_elapsed=3.438s timeout=1.000s
side_effect_survived_2_6_seconds=True
timeout_error=[timeout] the attempt ran past timeout=1s and was stopped
```

Runtime: Python 3.12.13, Django 5.2.17, psycopg 3.3.6, PostgreSQL 17.9. Consumer test callbacks are replaced in memory; worker execution is not mocked. A dedicated Heroku consumer-harness reproduction will be recorded before implementation changes. This is not a claim that a particular production incident has this cause.

### Expected behavior and regression gate

Consumer failure callbacks must not block or terminate the supervisor. Run this terminal callback work outside the supervisor, with a bounded lifetime, preserving committed terminal state and the documented best-effort callback contract.

Test stuck finalization while a sibling reaches its deadline. Require evidence of callback PID, sibling start and termination, absence of post-deadline effects, and supervisor survival. A callback that exits its process must also leave the supervisor alive. Do not count missing fixtures or a callback that never ran as a pass.

No library fix is included. First preserve the failing server regression, then fix upstream and rerun the same case against the release candidate.

### LINKED ISSUE #289: Unlock-before-commit stability test passes after a helper thread times out and rolls back
## Problem

`TestUnlockBeforeCommitWindow.test_second_transition_reads_stale_committed_state` reports success after its first transaction times out and rolls back in a helper thread. This does not prove the documented case where both callers commit.

Reviewed on v2.2.0, commit `a74391ca19dc3fc6719b31d6615ef521b2ceb7b2`. [Affected test](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/tests/stability/test_transaction_integration.py#L208-L264).

## Reproduction

Run the existing PostgreSQL/Redis stability test using `tests.settings_stability`:

```sh
python tests/manage.py test tests.stability.test_transaction_integration.TestUnlockBeforeCommitWindow --verbosity=2
```

T1 waits for `t2_probed` while retaining its PostgreSQL write lock. T2 sets that event only after `approve()` returns, but that call waits for T1's row lock. After ten seconds T1 raises `RuntimeError('T2 never probed')` and rolls back. T2 can then complete. The main test only checks that the thread stopped and the final state is `approved`; it never propagates T1's exception.

The September 22 review ran the combined background/stability subset on real PostgreSQL 17.9 and Redis 8.6.1 using both psycopg 3.3.6 and, in a separate environment, psycopg2 2.9.13 (Python 3.13.12, Django 5.2.17). Both logs contain:

```text
RuntimeError: T2 never probed
ok
...
Ran 332 tests
OK
```

This is a test defect, separate from the documented caller constraint that an outer transaction can outlive a synchronous transition's cache lock.

## Acceptance criteria

- Collect and fail on every helper-thread exception; assert all threads terminate.
- Signal the second caller reaching the relevant effect before its blocked state write, then allow T1 to commit.
- Assert both transactions commit and count the business effect, rather than accepting final state alone.
- Add a negative check showing that an intentionally failing helper makes this test fail.

No engine fix is proposed in this issue. An independently synchronized two-connection probe confirmed the documented duplicate-effect window with both transactions committed.

### LINKED ISSUE #290: Stability CI installs both PostgreSQL drivers and silently skips its claimed psycopg2 coverage
## Problem

The stability workflow says its PostgreSQL background run covers psycopg2, but it installs both drivers and Django selects psycopg3. This leaves the psycopg2 LISTEN/NOTIFY path without the continuous coverage claimed by the job.

Reviewed on v2.2.0, commit `a74391ca19dc3fc6719b31d6615ef521b2ceb7b2`.

- [Workflow installation and driver claim](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/.github/workflows/stability.yml#L54-L71): `pip install -e ".[dev]"`, then `pip install psycopg2-binary`.
- [The dev extra](https://github.com/Borderless360/django-logic/blob/a74391ca19dc3fc6719b31d6615ef521b2ceb7b2/pyproject.toml#L44-L53) already installs `psycopg[binary]>=3.2`.

## Reproduction

After those installation commands:

```sh
python -c 'from django.db.backends.postgresql.base import Database; print(Database.__name__, Database.__version__)'
```

Observed September 22 review environment: Django 5.2.17, psycopg 3.3.6 and psycopg2-binary 2.9.13 installed; selected backend driver **`psycopg 3.3.6`**.

The independent review also ran all 332 PostgreSQL background/stability tests in a psycopg2-only environment successfully. This issue is a missing CI gate, not evidence that 2.2.0's psycopg2 wake-up fix is broken. The separate thread-exception test problem remains present in both driver runs.

## Acceptance criteria

- Use separate, mutually exclusive psycopg2 and psycopg3 jobs or matrix environments.
- Assert the actual Django backend driver in each job before running tests; a wrong driver must fail setup.
- Run the real PostgreSQL notification and row-lock tests in both environments.
- Preserve the supported Python-version coverage and record selected driver/version in CI output.

No dependency or library changes have been made as part of this review.

emil-balashov:
Crash accounting and terminal callbacks could block the supervisor while other attempts exceeded their timeouts. This PR isolates those waits and keeps full workers responsive when attempts finish.

- Crash accounting takes a nonblocking PostgreSQL row lock and retains failed writes for retry.
- Safety nets run in one separate process per worker, limited to 60 seconds. Normal attempt deadlines remain active.
- Full workers use a bounded child wait with 10 ms exit checks. Finished slots accept queued work promptly. The wait returns for maintenance scheduling and does not accelerate database retries to the child polling rate.
- The transaction test propagates helper-thread failures and proves both callers commit. CI installs PostgreSQL drivers separately and verifies Django's selected driver.

Release: **2.2.1**. The runtime source matches verified commit `be49e5c03d3c84d2efeb0114e7b15f6f183542c3`. The final commit changes only package version and changelog heading.

Local validation: 592 SQLite tests passed, 35 backend-only skips. Each PostgreSQL driver passed all 344 background/stability tests. Counts overlap. The new backlog tests fail against the previous PR commit; the maintenance test also rejects unbounded full-worker waiting. Wheel and source archive builds pass.

The five-job consumer reproduction improved from 5.09 seconds on the previous PR commit to 0.23 seconds with the fix. Replacement gaps fell from about 980 ms to 10–16 ms. It uses real PostgreSQL, Redis and child processes.

Heroku verification passes on release **v123**, harness `4416f04`, with installed library commit `be49e5c` and matching source hashes. TC-47 fails on the previous candidate with actual exit 1 and passes on this commit with exit 0. All four replacement gaps improve from **0.92–1.00 seconds** to **0.13–0.16 seconds**. Every job completes once with zero errors.

The unchanged crash-accounting and callback regressions also exit 0. Accounting returns in **3.509 ms** during a six-second row lock. Sibling timeouts are recorded after **0.949** and **0.922 seconds**, with no late side effects. Independent reads confirm **34 durable events**, fixture cleanup, and all 224 prior completed messages.

Existing concurrency and notification checks pass: **8 passed, 0 failed**, including strict final fleet validation. All **15 CI jobs** and both automated reviews pass on this commit. The test app's original stopped formation is restored.

[Immutable results and evidence](https://github.com/Borderless360/django-logic-test/blob/07666f9f861a368145721944869ae469a6132e6e/docs/RESULTS.md) include the failing and passing runs, durable identities, source hashes and verification limits. The older 48 server check outcomes apply to `b06cacf`; this is a focused rerun.

The additional nonblocking review comments were verified locally and remain open: [retained accounting capacity #292](https://github.com/Borderless360/django-logic/issues/292), [silent lock deferrals #293](https://github.com/Borderless360/django-logic/issues/293), and [cleanup progress after a deadline #294](https://github.com/Borderless360/django-logic/issues/294). These are not claimed fixed or server-reproduced. The new maintenance regression requires two passes while an attempt remains active, addressing the recurring-maintenance coverage concern.

Reserve one extra process and database connection per worker, plus connections opened by callbacks. Callbacks remain best-effort. Required GV work must still have durable application recovery.

This PR prepares version 2.2.1 for the requested release. The release tag will trigger the existing publish workflow after merge. GV will receive separate PRs for the transformed copy and application review. Manual deployment-interruption, PgBouncer and other previously listed consumer scenarios remain outside the verification scope.

Fixes #287. Fixes #288. Fixes #289. Fixes #290.


django_logic/background/pull.py
```diff
@@ -501,17 +546,21 @@ def run_worker(
...
-        # for the whole life of the longest attempt.
-        full = len(attempts) >= concurrency
-        _harvest(attempts, block=full)
-        if time.monotonic() - last_safety_net >= SAFETY_NET_SECONDS:
-            _run_safety_nets()
-            last_safety_net = time.monotonic()
-        if claimed_any or full:
+        # A full worker still checks deadlines and schedules safety nets.
+        before_harvest = len(attempts)
+        _harvest(attempts, block=False)
```
emil-balashov, line 564: **Blocking.** This call plus the wait at the end of the loop adds up to `BUSY_POLL_SECONDS` after every attempt on a full worker. An attempt that ends sends no NOTIFY, so `_wait_for_work` runs to its limit before the loop reaps the child and claims the next row. On 2.2.0 a full worker blocked in `waitpid` and reaped at once.

Measured, draining 10 fast rows with `forever=True`:

| Concurrency | 2.2.0 | This branch | This branch with the change below |
|---|---|---|---|
| 1 | 0.28 s | 10.28 s | 1.42 s |
| 2 | 1.13 s | 5.26 s | 2.21 s |

A consumer that declares no `timeout=` pays this on every job. GV declares none and runs two slots per worker.

Change I tested:

```python
full = sum(
    attempt.pk is not None and not attempt.reaped
    for attempt in attempts.values()
) >= concurrency
before_harvest = len(attempts)
_harvest(attempts, block=full)
```

`_harvest` already polls at one-second steps while any child carries a deadline, so the safety-net child and every `timeout=` are still enforced, and it returns the moment a child exits. All 65 tests in `test_safety_net_isolation`, `test_pull_workers`, `test_accounting_contention`, `test_timeout_enforcement`, `test_safety_net` and `test_worker_holds_row` pass with it.

One residual stays: while the safety-net child is alive, about once a minute, a row attempt that ends waits up to one second. A SIGCHLD wake-up for `_wait_for_work` would remove that too.

django_logic/background/pull.py
```diff
@@ -478,14 +522,15 @@ def run_worker(
         ','.join(queues), concurrency,
     )
     attempts: dict[int, _Attempt] = {}
-    last_safety_net = 0.0
+    last_safety_net = None
     while True:
         claimed_any = False
         try:
-            while len(attempts) < concurrency:
+            while sum(attempt.pk is not None for attempt in attempts.values()) < concurrency:
```
emil-balashov, line 542: Not blocking. A reaped attempt whose accounting is deferred still counts toward `concurrency`. With two workers, the crashed worker's slot stays blocked for the whole replacement attempt. If the replacement hangs with no `timeout=`, the slot stays blocked until that attempt dies. 2.2.0 froze the whole worker in the same case, so this is not worse, but the slot is still lost.

Suggest: exclude reaped attempts from this count. Keep them in `exclude_pks`, so this worker does not claim its own crashed row before the error lands.

django_logic/background/pull.py
```diff
@@ -293,6 +308,12 @@ def _try_account(attempt: _Attempt) -> bool:
     try:
         _account(attempt, attempt.exit_code)
     except Exception as exc:
+        cause = exc.__cause__
+        if (getattr(cause, 'sqlstate', None)
+                or getattr(cause, 'pgcode', None)) == '55P03':
+            # Another worker holds the row. Retry without delaying the
+            # deadlines of the attempts this worker still supervises.
+            return False
```
emil-balashov, line 329: Not blocking. This path logs nothing. A row that stays locked for minutes, or for good, retries every second in silence, and the slot stays occupied. Suggest a warning after a few deferrals, with the pk and how long the write has waited.

django_logic/background/pull.py
```diff
@@ -63,6 +63,7 @@
 
 #: How often the loop runs the safety nets (stuck report, cleanup).
 SAFETY_NET_SECONDS = 60.0
+SAFETY_NET_TIMEOUT_SECONDS = 60.0
```
emil-balashov, line 69: Not blocking. `cleanup_completed_transitions` runs one DELETE in one transaction. A pass that needs more than 60 seconds is killed, rolls back, and starts again on the next pass, so it never completes. On 2.2.0 the same pass ran once, slowly, and finished.

The steady state is small, because the sweep runs every minute. The risk is after a long worker outage, or a first deploy onto a table with a backlog. Delete in batches of pks, so a killed pass keeps its progress.

tests/background/test_pull_workers.py
```diff
@@ -149,20 +149,35 @@ def test_no_error_is_recorded_on_a_row_another_worker_completed(self):
...
+                    while not marker.exists():
+                        if time.monotonic() >= limit:
+                            self.fail('The safety-net child did not record its pass.')
+                        time.sleep(0.01)
+            with patch('django_logic.background.pull._run_safety_nets', record_pass), \
+                    patch('django_logic.background.pull._start_attempt', start_and_wait_for_pass):
+                run_worker(_CRITICAL, forever=False)
+            # The child records shared evidence before the one-slot worker
+            # can finish the second row. Parent-only lists cannot record it.
+            self.assertGreaterEqual(int(marker.read_text()), 1)
```
emil-balashov, line 180: Not blocking. The old assertion required at least two safety-net passes during the drain. This one requires one pass while at least one row is still uncompleted. That is weaker. With `SAFETY_NET_SECONDS` patched to 0 and `forever=False`, a second pass starts after the second claim, so a per-pass marker could still prove two passes. If one pass is the intended contract now, say so in the test name or docstring.

emil-balashov (commented):
Review of `b06cacf` against 2.2.0 (`a74391c`). I ran both in separate worktrees against PostgreSQL 17 and Redis with psycopg3, and I ran the new tests of this PR against 2.2.0.

**Verdict.** The four fixes are correct, and their new tests reproduce the defects on 2.2.0. One regression blocks the merge: a full worker now waits up to one second after each attempt ends before it claims the next row. See the inline comment on `run_worker`.

**What I verified**

- #287: the new tests pass here. On 2.2.0 they fail as the issue predicts. The accounting write blocked for the whole lock hold, 4.0 s against the 2 s limit, and `_try_account` returned True where the deferred path expects False. The `55P03` check works on both drivers: psycopg3 exposes `sqlstate`, psycopg2 exposes `pgcode`, and Django chains the driver error as `__cause__`. On SQLite the lock request compiles to nothing and the write still runs. After a SIGKILL, the row lock was already free before the accounting write in 10 of 10 runs, so the `nowait` lock adds no retry delay on the normal kill path.
- #288: the new tests pass here. On 2.2.0, `test_callback_process_exit_does_not_stop_the_worker` ended the test runner itself with exit code 17, and `test_slow_callback_does_not_delay_a_sibling_timeout` found the callback pid equal to the supervisor pid.
- #289: the rewritten test passes here and on 2.2.0. The engine is unchanged, so that is the expected result. The negative check proves the helper thread propagates its error.
- #290: the psycopg2 job log shows `Django selected psycopg2 2.9.13`, 287 background tests and 53 stability tests, and every new test ran and passed. None was skipped.
- The linked results document at `58fcff7` carries the numbers in the PR body.

**Outside the diff**

- The module docstring of `django_logic/management/commands/dl_worker.py` still says the loop runs the safety nets. The loop now starts a child process for them.
- Environment note: under psycopg2 on macOS the forked child segfaults with exit code -11, on 2.2.0 and on this branch alike. Linux CI passes. This is not a defect of this PR.

